PR #5, opened 2023-08-27
Installation fixes

michalsn:
I came across some issues during the installation process. This PR is about to fix them.

* update composer.lock - otherwise the `composer install` command returns a message
> - Required package "michalsn/codeigniter-htmx" is in the lock file as "v1.2.1" but that does not satisfy your constraint "^1.2.2".

* add `writable` folder to the repo, to fix fatal error
> Fatal error: Uncaught CodeIgniter\Cache\Exceptions\CacheException: Cache unable to write to "/cache/". in /Users/michalsn/Sites/forum/vendor/codeigniter4/framework/system/Cache/Handlers/FileHandler.php:64

* update spark command from `php spark migrate` to `php spark migrate --all` which will also run Shield migrations

### DIFF
--- a/_docs/start.md
+++ b/_docs/start.md
@@ -23,7 +23,7 @@ Once the reposoitory has been installed on your local machine you need to get a 
 ```cli
 composer install  (only needed if you directly downloaded the files)
 npm install
-php spark migrate
+php spark migrate --all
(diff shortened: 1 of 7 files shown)


datamweb (approved):
I lost a lot of time last night while installing.
thank you!

PR #19, opened 2023-08-28
fix: Toolbar css

michalsn:
This PR fixes a problem with the `.tab` class that messes up the debug toolbar details view.


datamweb (approved):

lonnieezell:
Thanks!

PR #20, opened 2023-08-29
fix: Reduce the number of produced queries

michalsn:
This PR reduces the number of executed queries in the Discussions tab.

Now, we're dealing with n+1 problem. Luckily it was easily fixable.


datamweb (approved):

lonnieezell (approved):
Nice catch. Thanks!

I threw this together fairly quickly over a few days time, IIRC. I'm sure that's not the only place that could use optimization lol.

PR #25, opened 2023-08-30
fix: load Alpinejs by default

### LINKED ISSUE #24: Alpine.js is not loaded by default
Why doesn't it work?
https://github.com/lonnieezell/forum-example/blob/767790bc26670695953aa9902b4a05cd1d8d4c1b/package.json#L11

https://github.com/lonnieezell/forum-example/blob/767790bc26670695953aa9902b4a05cd1d8d4c1b/themes/default/js/app.js#L3

https://github.com/lonnieezell/forum-example/blob/767790bc26670695953aa9902b4a05cd1d8d4c1b/package.json#L25

So I expect the following code to work in any view without problems, but it doesn't.

```html
    <div x-data="{ open: false }">
        <button @click="open = true">Expand</button>

        <span x-show="open">
            Content...
        </span>
    </div>
```
I couldn't figure out what was wrong.

**Note:**
The following code works in any view.
```diff
+<script src="//unpkg.com/alpinejs" defer></script>
    <div x-data="{ open: false }">
        <button @click="open = true">Expand</button>

        <span x-show="open">
            Content...
        </span>
    </div>
```

datamweb:
fix #24

### DIFF
--- a/themes/default/js/app.js
+++ b/themes/default/js/app.js
@@ -1,3 +1,7 @@
 import "htmx.org";
 import "htmx.org/dist/ext/ajax-header";
-import "alpinejs";
+import Alpine from 'alpinejs';
+ 
+window.Alpine = Alpine;
+ 
+Alpine.start();


lonnieezell (approved):
Thank you!

PR #26, opened 2023-08-30
feat: Members list

### LINKED ISSUE #13: Visitors should be able to see members of the site.
A new "Members" page should be present that displays a paginated list of all users within the system. 

They should be presented with a search for this page only that searches and filters the list of users by the following criteria: 
- username
- role (admin, moderator especially)
- new users (with 1 or less posts)
- country

You should be able to sort the list of users by : 
- number of posts
- username, asc
- most recent posts, desc

michalsn:
Closes #13 

This is my first encounter with Tailwind without ready-made UI components. It's hard. So far I don't grasp DaisyUI at all.
In addition, not all examples look like in the documentation.... but maybe it's the fault of the configuration... IDK. Sorry for bragging about my poor CSS skills 😅 

Not sure if this member page layout will stay for long, but it's a start.

Anyway... any suggestions for changes are very welcome.

### DIFF
--- a/app/Cells/MemberList.php
+++ b/app/Cells/MemberList.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace App\Cells;
+
+use App\Concerns\HasTableSearchComponent;
+use App\Models\UserModel;
+use CodeIgniter\Pager\Pager;
+use CodeIgniter\View\Cells\Cell;
+
+class MemberList extends Cell
+{
+    use HasTableSearchComponent;
+
+    protected string $view = 'member_list_cell';
+
+    protected array $members;
+    protected Pager $pager;
+
+    public function mount()
+    {
+        $this->defineTableSearchComponent();
+
+        helper('form');
+
+        $model = model(UserModel::class);
+
+        $this->members = $model->searchMembers(
+            $this->search, $this->page, $this->perPage, $this->sortColumn, $this->sortDirection
+        );
+
+        $this->pager = $model->pager->setPath($this->baseURL);
+    }
+
+    protected function getMembersProperty(): array
+    {
+        return $this->members;
+    }
+
+    protected function validRoles(): array
+    {
+        return array_combine($this->validSearch['role'], $this->validSearch['role']);
+    }
+
+    protected function validTypes(): array
+    {
+        return array_combine($this->validSearch['type'], $this->validSearch['type']);
+    }
+
+    private function defineTableSearchComponent(): void
+    {
+        $this->defineBaseURL('members');
+
+        $this->defineSortColumns(
+            service('request')->getGet('sortColumn') ?? 'username',
+            ['in_list' => ['username', 'role', 'count', 'country', 'last_active']]
+        );
+
+        $this->defineSortDirections(
+            service('request')->getGet('sortDirection') ?? 'asc',
+            ['in_list' => ['asc', 'desc']]
+        );
+
+        $this->definePerPage(
+            service('request')->getGet('perPage') ?? 5,
+            ['in_list' => [5, 10, 20]]
+        );
+
+        $this->definePage(
+            service('request')->getGet('page') ?? 1,
+            ['is_natural_no_zero']
+        );
+
+        $search = service('request')->getGet('search');
+        $this->defineSearch([
+            'user' => [
+                $search['user'] ?? '',
+                ['string', 'max_length[10]']
+            ],
+            'country' => [
+                $search['country'] ?? '',
+                ['string', 'max_length[10]']
+            ],
+            'role' => [
+                $search['role'] ?? 'all',
+                ['in_list' => array_merge(['all'], array_keys(setting('AuthGroups.groups')))]
+
+            ],
+            'type' => [
+                $search['type'] ?? 'all',
+                ['in_list' => ['all', 'new']]
+            ],
+        ]);
+    }
+}
--- a/app/Cells/MemberList.php
+++ b/app/Cells/MemberList.php
@@ -38,12 +38,12 @@ class MemberList extends Cell
 
     protected function validRoles(): array
     {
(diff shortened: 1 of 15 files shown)


lonnieezell:
I'll review more later, but I thought I'd try to help clear up a couple things a bit. 

1) DaisyUI is built on top of Tailwind, and provides Bootstrap like classes (`.card`, `.alert`, `.alert-info`, etc) that can you use to build UI on, since Tailwind doesn't provide anything but utility classes. It still allows you to further refine the elements with Tailwind classes in the HTML, though, so you gain Bootstrap's ease of use with Tailwind's flexibility and power. 
2) We can always create our own base classes in `themes/default/css/app.scss` to use like normal.

lonnieezell:
Sidenote: @michalsn I know you've used Cells quite a bit in other work so I have a question. Have you found it helpful to do things like the Posts list, or this member list as a cell? I'm on the fence about whether it makes sense to do that (like I currently am) or it's just simpler to do it the old way and utilize cells for interactive templates, or common elements (like a category list). If they were live cells I know what my answer would be, just curious what you've found using it in other projects.

app/Models/UserModel.php
```diff
@@ -2,10 +2,63 @@
...
+        'active',
+        'last_active',
+        'handed',
+        'thread_count',
+        'post_count',
+        'avatar',
+        'country',
+        'timezone',
+        'deleted_at',
+    ];
```
datamweb:
```suggestion
    protected function initialize(): void
    {
        parent::initialize();

        // Merge properties with parent
        $this->allowedFields = array_merge($this->allowedFields, [
            'handed',
            'thread_count',
            'post_count',
            'avatar',
            'country',
            'timezone',
        ]);
       
    }
```

themes/default/master.php
```diff
@@ -13,6 +13,7 @@
     <link rel="stylesheet" href="<?= base_url('css/app.css') ?>">
     <script src="https://kit.fontawesome.com/9ce245a629.js" crossorigin="anonymous"></script>
     <?= $this->renderSection('styles') ?>
+    <script src="<?= base_url('js/app.js') ?>"></script>
```
datamweb, line 16: It is usually recommended to load file **JS** at the end of the HTML. I was curious to see if there is a special reason for moving?

datamweb:
> This is my first encounter with Tailwind without ready-made UI components. It's hard. So far I don't grasp DaisyUI at all. In addition, not all examples look like in the documentation.... but maybe it's the fault of the configuration... IDK. Sorry for bragging about my poor CSS skills 😅


Your situation is much better than mine, I'm the first to touch Tailwind, DaisyUI, AlpineJs.

> 1. DaisyUI is built on top of Tailwind, and provides Bootstrap like classes (`.card`, `.alert`, `.alert-info`, etc) that can you use to build UI on, since Tailwind doesn't provide anything but utility classes. It still allows you to further refine the elements with Tailwind classes in the HTML, though, so you gain Bootstrap's ease of use with Tailwind's flexibility and power.

My lack of familiarity with this tool slowed me down. But I think I should have started somewhere. I think that under these conditions, the speed of development will slow down. I am looking to help.

Demo:
Screenshot 2023-08-30 220543

michalsn:
@lonnieezell Here I followed the convention I saw in the project already. That's why I created a cell for the members list.

From my experience, I tend to use cells for widgets or common elements (to easily display them on different occasions). They are perfect especially if we don't have to save the state of the component in the URL.

In general, I follow the rule: if the actions performed by the user should be reflected by the URL address change, I choose the controller. Everything else can be a component. 

This doesn't mean that components can't be interactive, but that the state of the component (e.g., like the current page in the table) isn't necessarily restorable via URL (the one displayed in the browser bar).

app/Models/UserModel.php
michalsn: Perfect, thanks!

themes/default/master.php
michalsn: With htmx we usually want the SPA-like experience. The idea is to load the required (main) JS only once. Then, between the requests, htmx will replace only the `body` part of the page. Putting `app.js` in the header will prevent loading this file again on every page request.

michalsn:
@datamweb I know Tailwind only briefly. I already know I will suck at the UI part... but on the bright side, it's an opportunity to learn Tailwind a bit.

themes/default/master.php
datamweb: Thank you!

app/Cells/MemberList.php
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    protected function getMembersProperty(): array
+    {
+        return $this->members;
+    }
+
+    protected function validRoles(): array
+    {
+        return array_combine($this->validSearch['role'], $this->validSearch['role']);
```
datamweb: @michalsn 
`$this->validSearch['role']` produces the following string. This is not good, we need the **role title**.
```php
all => string (3) "all"
⇄superadmin => string (10) "superadmin"
⇄admin => string (5) "admin"
⇄developer => string (9) "developer"
⇄user => string (4) "user"
⇄beta => string (4) "beta"
```

**Hard Code**
```php
        return array_combine($this->validSearch['role'], [
            'all' =>  "all",
            'superadmin' => setting('AuthGroups.groups')['superadmin']['title'],
            'admin'      =>  setting('AuthGroups.groups')['admin']['title'],
            'developer'  => setting('AuthGroups.groups')['developer']['title'],
            'user'       =>  setting('AuthGroups.groups')['user']['title'],
            'beta'       =>  setting('AuthGroups.groups')['beta']['title'],
        ]);
```

**Befor**

before

**After**

after

michalsn: It was on my to-do list for later, but since you mentioned it... I added it now. Thanks.

datamweb: Thank you!

PR #29, opened 2023-09-01
feat: Member list - with controller

### LINKED ISSUE #13: Visitors should be able to see members of the site.

michalsn:
An alternative version to #26 - this time using a more standard approach: the controller.

Closes #13

### DIFF
--- a/app/Concerns/ImpactsForumCounts.php
+++ b/app/Concerns/ImpactsForumCounts.php
@@ -28,7 +28,7 @@ trait ImpactsForumCounts
 
         $forumModel->save($forum);
 
-        // Decrement User thread count
+        // Increment User thread count
         $user = $userModel->find($thread->author_id);
         $user->thread_count++;
 
--- a/app/Concerns/ImpactsForumCounts.php
+++ b/app/Concerns/ImpactsForumCounts.php
@@ -5,6 +5,7 @@ namespace App\Concerns;
 use App\Entities\Forum;
 use App\Models\ForumModel;
 use App\Models\ThreadModel;
+use App\Models\UserModel;
 
 trait ImpactsForumCounts
 {
@@ -18,10 +19,20 @@ trait ImpactsForumCounts
         }
 
         $thread = $this->find($data['id']);
-        $forum = model(ForumModel::class)->find($thread->forum_id);
+        $forumModel = model(ForumModel::class);
+        $userModel = model(UserModel::class);
+
+        // Increment Forum thread count
+        $forum = $forumModel->find($thread->forum_id);
         $forum->thread_count++;
 
-        model(ForumModel::class)->save($forum);
+        $forumModel->save($forum);
+
+        // Decrement User thread count
+        $user = $userModel->find($thread->author_id);
+        $user->thread_count++;
+
+        $userModel->save($user);
 
         return $data;
     }
@@ -36,10 +47,21 @@ trait ImpactsForumCounts
         }
 
         $thread = $this->find($data['id']);
-        $forum = model(ForumModel::class)->find($thread->forum_id);
+
+        $forumModel = model(ForumModel::class);
+        $userModel = model(UserModel::class);
+
+        // Decrement Forum thread count
+        $forum = $forumModel->find($thread->forum_id);
         $forum->thread_count--;
 
-        model(ForumModel::class)->save($forum);
+        $forumModel->save($forum);
+
+        // Decrement User thread count
+        $user = $userModel->find($thread->author_id);
+        $user->thread_count--;
+
+        $userModel->save($user);
 
         return $data;
     }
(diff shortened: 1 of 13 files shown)


app/Concerns/ImpactsForumCounts.php
```diff
@@ -18,10 +19,20 @@ protected function incrementThreadCount(array $data)
...
+        $userModel = model(UserModel::class);
+
+        // Increment Forum thread count
+        $forum = $forumModel->find($thread->forum_id);
         $forum->thread_count++;
 
-        model(ForumModel::class)->save($forum);
+        $forumModel->save($forum);
+
+        // Decrement User thread count
```
lonnieezell:
```suggestion
        // Increment User thread count
```

app/Models/UserModel.php
```diff
@@ -2,10 +2,85 @@
...
+        $selects = [
+            'users.*',
+            '(thread_count + post_count) AS count',
+        ];
+
+        $results = $this
+            ->withIdentities()
+            ->select(implode(', ', $selects))
+            ->join('auth_groups_users', 'auth_groups_users.user_id = users.id', 'left')
+            ->when(isset($search['username']) && $search['username'] !== '', static function ($query) use ($search) {
```
lonnieezell: Not saying you need to change it, but newer versions of PHP 8 allow new function types that are scoped to the outer function, like: 

```suggestion
            ->when(isset($search['username']) && $search['username'] !== '', fn ($query) => $query->like('users.username', $search['username'], 'both');
```

Or something like that.

lonnieezell (approved):
A couple of notes in there, but I like going this way instead of the cell better. I think it's more straight-forward and a better practice for people looking at our code. 

Thanks!

app/Models/UserModel.php
michalsn: Will do. I've recently spent some time with older PHP 7 projects, so the new syntax doesn't come naturally.

app/Views/members/list.php
```diff
@@ -0,0 +1,94 @@
...
+                                </div>
+                                <div class="px-4">
+                                    <div class="font-bold cursor-pointer" hx-get="<?= esc($member->link(), 'attr') ?>"><?= esc($member->username); ?></div>
+                                    <div class="text-sm opacity-50"><?= is_array($member->roles) ? implode(', ', $member->roles) : $member->roles; ?></div>
+                                </div>
+                            </div>
+                        </td>
+                        <td><?= $member->count; ?></td>
+                        <td><?= esc($member->country); ?></td>
+                        <td><?= $member->last_active->humanize() ?? 'Never'; ?></td>
```
datamweb:
```suggestion
                        <td><?=($member->last_active !== null) ? $member->last_active->humanize() : 'Never';?></td>
```

app/Controllers/Members/MemberController.php
```diff
@@ -0,0 +1,71 @@
...
+            'sortDirection'   => ['in_list[asc,desc]'],
+        ];
+
+        if (! $this->validateData($table, $rules)) {
+            throw new InvalidArgumentException(implode(PHP_EOL, $this->validator->getErrors()));
+        }
+
+        helper('form');
+
+        $model = model(UserModel::class);
```
datamweb:
```suggestion
        $userModel= model(UserModel::class);
```

app/Controllers/Members/MemberController.php
```diff
@@ -0,0 +1,71 @@
...
+
+        $data = [
+            'members' => $model->searchMembers(
```
datamweb:
```suggestion
            'members' => $userModel->searchMembers(
```

app/Controllers/Members/MemberController.php
```diff
@@ -0,0 +1,71 @@
...
+                $table['search'], $table['page'], $table['perPage'], $table['sortColumn'], $table['sortDirection']
+            )
+        ];
+
+        $table['dropdowns'] = [
+            'role' => array_combine($roleKeys, array_merge(['All'], array_column(setting('AuthGroups.groups'), 'title'))),
+            'type' => ['all' => 'All users', 'new' => 'New users'],
+        ];
+
+        $table['pager']  = $model->pager;
```
datamweb:
```suggestion
        $table['pager']  = $userModel->pager;
```

datamweb (approved):
@michalsn thank you!
LGTM!

michalsn:
Thanks, everyone.

PR #31, opened 2023-09-06
feat: create a thread and post

michalsn:
This PR adds an option to create threads and posts.

I'm not sure how we'll handle the permission check, but for now, I've implemented the check in the controller. Although lack of permissions results in `dd('Unauthenticated')` 😅  At some point we will have to change this.

I have split the code into different controllers and I'm planning to do so also with `DiscussionController` (but later).


lonnieezell (approved):

datamweb:
@michalsn , I think it is better to add this button(start ...) in **themes\default\master.php** so that it is present on all pages.

In addition, we can add a shortcut key(e.g. shift+N+D) to jump .

see [here ](https://htmx.org/examples/keyboard-shortcuts/).


Screenshot 2023-09-07 204327

michalsn:
You might be right. However, I'm not entirely sure it should be always available - e.g. a Members page.

PR #33, opened 2023-09-07
feat: add a reply to a post

### LINKED ISSUE #9: Logged in users should be able to create a new post
A new button should be present in the sidebar titled "Start a Discussion". This is only available when the user is logged in. When clicked, this should present a form for the user to start a new discussion. 

The form should appear on a new page. 

The following fields should be available: 

- Title
- Message - in a Markdown/BBCode-supporting editor 
- Save button

> Note: new posts should support Markdown, but old posts that we would import would need to support BBCode. Viewing a post already supports both languages.

michalsn:
This PR adds the ability to reply to posts

Closes #9


app/Entities/Post.php
```diff
@@ -34,9 +34,7 @@ public function link()
 
     public function setReplyTo(?string $value = null)
     {
-        if (empty($value)) {
-            $this->attributes['reply_to'] = null;
-        }
+        $this->attributes['reply_to'] = $value ?: null;
```
datamweb, line 37:
```suggestion
        $this->attributes['reply_to'] ??= $value;
```

michalsn: The goal here is to convert an empty string to `null`.

datamweb:
I am Reply **1 post** , but the answer is showing under **2 POST**. I expect **Reply to 1 post** to appear below 1 Post.

Screenshot 2023-09-07 224845

michalsn:
The way posts are displayed has not been changed. They are displayed in a linear fashion. I don't know if we will change it or not.

datamweb:
Good example on GitHub Discussions. Answers are linear until they are posted.
If a new post is created, the answer should be written below the new post.

see https://github.com/codeigniter4/shield/discussions/765

michalsn:
If this is to be changed, we need to determine how it should look/work
* The way to display nested posts (should it look like GitHub Discussions?)
* Whether to show all nested posts at once, add pagination, or a "more" button.
* Number of levels - because we can't nest posts indefinitely.

Probably worth addressing in a separate issue. For now, we keep all the necessary information, so we can figure out how to display it later.

datamweb (approved):
@michalsn thank you!

michalsn:
One important thing to note... the goal is also to import data from the old forum at some point.... so posts there may have a multi-level structure. Although I haven't checked what the DB schema looks like in MyBB.

lonnieezell:
The MyBB posts shema doesn't seem to link to other posts as a parent. I didn't see any other tables that I would imply otherwise either. 

Screenshot 2023-09-07 at 10 41 45 PM

I do think we should discuss the threading of posts and decide how necessary it is. If we did, I don't think we'd need to thread it deeper than 1 level, kind of like Slack. Especially since forum posts are usually wordy affairs :)

lonnieezell (approved):

lonnieezell:
I approved, and am going to merge, but we should probably start building up a test suite and get some CI going on here since this is no longer just me slinging code over a couple weekends for the fun of it :)

michalsn:
> I approved, and am going to merge, but we should probably start building up a test suite and get some CI going on here since this is no longer just me slinging code over a couple weekends for the fun of it :)

Here comes a thief of joy, the smiler reducer, a.k.a. the testing suite, ready to turn your code party into a debugging disco  😆 

Yeah... you're right. Thought it was fun without them.

michalsn:
@lonnieezell Is the `replyto` column in the screenshot associated with the user?

lonnieezell:
> @lonnieezell Is the `replyto` column in the screenshot associated with the user?

Sorry - missed this. I have no idea :) They don't have foreign keys setup. I would guess it was replying to another post, but not really sure. 

I could get you a copy of the SQL dump if you'd like it. It's only 96mb, zipped, and like 355mb unzipped :)

michalsn:
Nah, that's ok. If it's the post_id we're responding to, it would potentially create multi-level threading... old posts should be displayed in a linear fashion, which makes more sense since they are displayed that way now.

PR #35, opened 2023-09-08
Categories

### LINKED ISSUE #23: Users should see a list of all thread categories in the sidebar
- Posts should have categories. These are essentially the same as the existing forums for now. Renaming and refactoring is probably enough here. 
- Posts can only belong in one category (or sub-category)
- A category is required
- A cell should be created that lists the categories and its children in a nested fashion. 
- The cell must allow a category ID to be specified as the active one. 
- The active category must be visible and highlighted on the list. 
- If the category is a sub-category, the parent must be open so it can be seen. 
- Clicking the category/sub-category names will take the user to a page that lists the discussions, but only within the given category.

lonnieezell:
- refactor all mentions of `forums` into `categories` to transition to a lighter, less rigid forum system, more akin to Flarum. 
- Provide a Cell to list all categories in the system in a collapsible tree. 
- rip generated assets out of the system for cleaner merges. 

Fixes #23


michalsn (approved):
Ok, I thought `forums` and `categories` were different things and was wondering why we need both... now I get it.

michalsn:
We're missing the migration file that will rename the table to `categories`.

datamweb:
It looks like **categories** should be displayed in the sidebar by default, but it isn't.
I expect it to be displayed as below by default:

Screenshot 2023-09-08 214908

lonnieezell:
I'm not convinced it should be in the master template, if that's what you mean. I have a feeling that the pages will be varied enough that it doesn't make sense to have those on all pages. Things like member lists, profile pages, etc, don't need them.

lonnieezell:
@michalsn I didn't do a new migration since this project is so young and there's been no releases, yet. I updated the original file so we'll need to start from a clean slate and run all of the migrations again. It also updates column names in other tables: 

https://github.com/lonnieezell/forum-example/pull/35/files#diff-9318acdc3abb35b760632307636ee15a475ed7e7baa2e19feb9754f804919a2d

PR #37, opened 2023-09-08
fix: use `csrf` in form

datamweb:


michalsn:
I believe `form_open()` handles this automatically.

datamweb:
`form_open()`  handles this automatically when filter `'csrf'` below is active. There are currently `// 'csrf',`.

Am I missing something?
https://github.com/lonnieezell/forum-example/blob/eb3a12cc5208f812b993a8512efadc1c602e3ad7/app/Config/Filters.php#L36-L47

michalsn:
If `csrf` filter is disabled, then even if we add `csrf_field()` to the form, it won't be verified. So, there is no point in adding this field manually.

datamweb:
Ah, so we need to enable the csrf filter in app/Config/Filters.php.

michalsn:
Yes, that's all we need - nothing more.

PR #39, opened 2023-09-08
chore: add dependabot

datamweb:
Hello to the utility.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,36 @@
+version: 2
+updates:
+  - package-ecosystem: composer
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "08:30"
+      timezone: "America/Los_Angeles"
+    open-pull-requests-limit: 5
+    target-branch: "develop"
+    labels:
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -7,16 +7,16 @@
   "require": {
     "php": "^7.4 || ^8.0",
     "ext-xsl" : "*",
-    "codeigniter4/framework": "^4.4",
-    "codeigniter4/shield": "^1.0@beta",
-    "league/commonmark": "^2.3",
-    "michalsn/codeigniter-htmx": "^1.2.2",
-    "s9e/text-formatter": "^2.13"
+    "codeigniter4/framework": "4.4.1",
```
lonnieezell: I don't see the need to bake these to a path release version. I'd rather rely on semantic versioning and allow things to be updated when it should be safe to do so.

datamweb: I removed the commit for it, hope I got it right.

lonnieezell:
Looks good, thanks!

michalsn (approved):

PR #41, opened 2023-09-09
feat: edit thread and post

### LINKED ISSUE #10: Users should be able to edit a discussion
As a user I should be able to edit an existing post. I can always edit my own post. Otherwise, it should be based on a permission check whether the edit link even shows up or not. 

When clicked, this should provide the option to edit the post in place and save the changes.

michalsn:
This PR adds the option to edit posts and threads. When the `category_id` is changed (while editing a thread), we recalculate all statistics for the new and old category to keep the data in sync.

Closes #10

I know... still no tests. If no one else has started this before, the next step will probably be to set up PHPUnit and write some tests.


michalsn:
I had to temporarily disable the CSRF filter due to a problem with checking the raw body in the PUT request type. I hope this will be resolved in version 4.4.2. An improvement is on the way: https://github.com/codeigniter4/CodeIgniter4/pull/7915

datamweb (commented):
Can you consider this?
Screenshot 2023-09-10 211148

datamweb:
> the next step will probably be to set up PHPUnit and write some tests.

I agree and I also don't know what code quality control tool the team wants to use. Just a PHPUnit ? Or other tools(rector,phpstand ...) like shield?

datamweb:
Create a post with the following text:
````
```
echo 'Hi';
```
````
Then edit it, is displayed as below:
````
```
echo &#039;Hi&#039;;
```
````

**Comment by michalsn** [comment]
> Can you consider this?

What do you mean? Post and thread editing is done inline, so the rest of the interface stays untouched.

> Just a PHPUnit ? Or other tools(rector,phpstand ...) like shield?

I already started doing it. I have 5 things: PHPStan, Psalm, Rector, PHP CS Fixer, and PHPUnit.

**Comment by michalsn** [comment]
The problem with escaping entities should be fixed. Thanks!

**Comment by datamweb** [comment]
> What do you mean? Post and thread editing is done inline, so the rest of the interface stays untouched.

Sorry, I meant padding.

**Comment by michalsn** [comment]
Done.

**Review by datamweb — APPROVED** [review]
LGTM!
Thank you!

File: app/Controllers/Discussions/PostController.php [hunk 1]
```diff
@@ -54,7 +59,47 @@ public function create(int $threadId, ?int $postId = null)
...
+
+        $post = $postModel->find($postId);
+
+        if (empty($post) ||
+            ($post->author_id !== user_id() &&
+                ! auth()->user()?->can('posts.edit'))) {
+            dd('Unauthorized');
+        }
+
+        if ($this->request->is('put')) {
```
lonnieezell, line 80: I'm not sure how well `put` will work here if we allow uploading images, but we can figure that out when we get there.

lonnieezell (approved):

app/Controllers/Discussions/PostController.php
michalsn: We will probably do it with a separate request, and then use markdown to embed the file. At least this is how I imagine it now.

lonnieezell: Makes perfect sense to me.

PR #46, opened 2023-09-11
feat: add ci and code fixes

michalsn:
This PR adds support for PHPStan, Psalm, Rector, PHP CS Fixer, and PHPUnit.

So far there were not many problems, but I had to ignore the `TextFormatter` class because I had no idea how to handle PHPStan errors.... (other than individually, which didn't make much sense to me). Maybe there is a way to improve it, but for now, it should work.


app/Libraries/TextFormatter.php
```diff
@@ -67,19 +68,20 @@ private function markdownParser()
...
 
     /**
      * Returns the Markdown renderer
+     *
      * @return mixed
      */
     private function markdownRenderer()
     {
-        if (! is_null($this->markdownRenderer)) {
+        if (null !== $this->markdownRenderer) {
```
lonnieezell: Why does our Rector force this one? This is bringing back Yoda comparisons. Whatever.

lonnieezell (approved):
Thanks for taking the time to do this. I still think we might have too many checks. The first time Rector says I can't merge because I don't have generics in my docblock I'm ripping it out :)

app/Libraries/TextFormatter.php
michalsn: I was a bit surprised too, because it's not how I would write the code, but I have no stamina today for searching how to change this.

michalsn:
> Thanks for taking the time to do this. I still think we might have too many checks. The first time Rector says I can't merge because I don't have generics in my docblock I'm ripping it out :)

I'm with you.

I really hope I won't try to cut myself the next time I try to send a PR 😅

phpstan.neon.dist
```diff
@@ -0,0 +1,37 @@
+parameters:
+	tmpDir: build/phpstan
+	level: 5
```
datamweb, line 3: Codeigniter has recently been launched to level 6.

https://github.com/datamweb/CodeIgniter4/blob/bdbb77566563e1bc5c5668fcb3a58b202946e143/phpstan.neon.dist#L17

michalsn: Well... I'm afraid that level 6 will flood us with errors 😜 

Let's keep it and see how all this is going to work for us in this project.

datamweb: Ok, right now we have to say hello to the brain.

datamweb (approved):

app/Libraries/TextFormatter.php
michalsn: Fixed... it bothered me. The CodeIgniter coding standard had this rule, which is now overridden.

lonnieezell: Awesome. Thank you!

PR #47, opened 2023-09-12
chore: add Cells to coverage exclude

datamweb:
```
Generating code coverage report in Clover XML format ... <div class="category-list">
    <h2>Categories</h2>

            <p>There are no categories to display.</p>
    </div>
<div class="discussion-list">
    <div class="mt-4 my-6 flex justify-between">
        <form hx-get="https://example.com/" hx-target="#discussion-list">
            <select name="type" class="select w-auto" onChange="this.form.submit()">
                <option value="recent-threads"
                    Undefined variable $listType
Error: Process completed with exit code 2.

```
https://github.com/lonnieezell/forum-example/actions/runs/6151297508/job/16691053832


datamweb:
```
 Directory of P:\MyGitHubWork\codeigniter4-forum-example\app\Cells

09/12/2023  07:24 AM               748 CategoryListCell.php
09/12/2023  07:08 AM             1,471 category_list.php
09/12/2023  07:08 AM               573 DiscussionList.php
08/26/2023  09:55 PM             1,509 discussion_list_cell.php
09/12/2023  07:08 AM               741 ShowThread.php
09/12/2023  07:08 AM             1,966 show_thread_cell.php
               6 File(s)          7,008 bytes
               0 Dir(s)  18,252,775,424 bytes free
```

show_thread_**cell**.php
discussion_list_**cell**.php

Why **category_list.php** why not  **category_list_cell.php**?
If we change the file name to category_list_**cell**.php, we can proceed as follows:
```
<directory suffix="_cell.php">./app/Cells</directory>
```

michalsn:
It seems to me that this is the result of a new naming convention, created when using `make:cell` command.

If the suffix will work better for us, then just rename the view cell name.

michalsn:
I think you also have to register the custom view cell name, just like we did in other cells.

datamweb:
> I think you also have to register the custom view cell name, just like we did in other cells.

I don't know if this is really necessary, but I did it in a separate commit, if it is not needed, we can go back.
Thanks.

michalsn (approved):

PR #52, opened 2023-09-13
test: members

michalsn:
This PR adds tests for MemberController. 

It also increases the time limit for slow tests. I believe it's reasonable since we work with the database all the time.


lonnieezell (approved):
I agree on the time limit.

PR #56, opened 2023-09-14
code cleanup

michalsn:
This PR makes a code cleanup for Policies.

The `phpstan/phpstan-mockery` package is added to make PHPStan happy when working with Mockery.

Not sure how long we gonna stick to all these checks, but I have to admit (besides these silly complaints about types) that it makes code cleaner.

### DIFF
--- a/app/Libraries/Policies/Policy.php
+++ b/app/Libraries/Policies/Policy.php
@@ -1,10 +1,10 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
 
 namespace App\Libraries\Policies;
 
 use App\Entities\User;
-use InvalidArgumentException;
-use RuntimeException;
 
 /**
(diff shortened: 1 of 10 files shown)


app/Libraries/Policies/Policy.php
```diff
@@ -70,7 +70,7 @@ public function deny(string $message, int $status = 403)
      */
     public function withPolicy(PolicyInterface $policy): self
     {
-        $this->policies[get_class($policy)] = $policy;
+        $this->policies[$policy::class] = $policy;
```
lonnieezell, line 73: Interesting. I would have never thought to do it this way. Neat.

lonnieezell (approved):
thanks for doing this.

app/Libraries/Policies/Policy.php
michalsn: Me either.

PR #57, opened 2023-09-15
feat: post threading

michalsn:
This PR adds support for post-threading as described in https://github.com/lonnieezell/forum-example/discussions/36

Changes involve also:
* small changes to the `Policy` class
* and more tests for controllers


app/Models/PostModel.php
```diff
@@ -39,16 +39,107 @@ public function visible()
...
+    }
+
+    /**
+     * Get latest replies for posts by given IDs.
+     */
+    public function getReplies(array $postIds, int $limit): array
+    {
+        // Used to get last replies using rank
+        $subQuery = $this->builder()
+            ->select('*, ROW_NUMBER() OVER (PARTITION BY reply_to ORDER BY created_at DESC) AS reply_rank')
```
lonnieezell, line 97: This reminds me that I need to go study my SQL some more. I think I could have really used this a couple of days ago at work :)

app/Policies/PostPolicy.php
```diff
@@ -0,0 +1,50 @@
...
+use App\Libraries\Policies\PolicyInterface;
+
+class PostPolicy implements PolicyInterface
+{
+    /**
+     * Determines if the current user can create a new thread.
+     */
+    public function create(): bool
+    {
+        return true;
```
lonnieezell: I think my sample file may have been crap. We wouldn't actually want to return `true` for all of these unless we want guests and all other roles to be able to post. I can think of 3 ways to handle this: 

1)  The methods should accept the user instance and check that with standard Shield stuff. 
2) Completely remove the methods that don't have custom checks since they should use the standard Shield call at that point. 
3) Use the `before()` method to do a standard Shield permission check unless it's the `edit` method since that has additional needs. 

I'd opt for #2 as the simplest and cleanest personally.

lonnieezell (changes requested):
Looks great except the comment I left on the PostPolicy.

michalsn:
Hmm... seems like `PolicyTest` is a bit unstable. We run our tests in the random order. I think that using `$this->resetServices()` (in `setUp` method) would help, but when I tried that, I got an error exception
```
ErrorException: ini_set(): Session ini settings cannot be changed after headers have already been sent
```
Not sure how to handle this. It's not a problem when we run tests individually: 
```
composer test tests/Policies/PolicyTest.php
```
but with:
```
composer test
```
it randomly fails on `PolicyTest`. Any thoughts on how to fix this?

app/Policies/PostPolicy.php
michalsn: Done.

datamweb:
With `composer test`:
```console
PS P:\MyGitHubWork\codeigniter4-forum-example> composer test
> phpunit
PHPUnit 9.6.12 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.5 with Xdebug 3.1.4
Configuration: P:\MyGitHubWork\codeigniter4-forum-example\phpunit.xml.dist
Random Seed:   1694842519

.....................................                             37 / 37 (100%)

Nexus\PHPUnit\Extension\Tachycardia identified this sole slow test:
+-------------------------------------------------------------------------+---------------+-------------+
| Test Case                                                               | Time Consumed | Time Limit  |
+-------------------------------------------------------------------------+---------------+-------------+
| Tests\\Controllers\\DiscussionControllerTest::testShowDiscussionsThread | 00:00:01.08   | 00:00:01.00 |
+-------------------------------------------------------------------------+---------------+-------------+


Time: 00:27.963, Memory: 54.00 MB

OK (37 tests, 140 assertions)

Generating code coverage report in PHP format ... done [00:00.002]

Generating code coverage report in Clover XML format ... done [00:00.121]

Generating code coverage report in HTML format ... done [00:00.563]
```

With `php vendor/bin/phpunit`:
```console
PS P:\MyGitHubWork\codeigniter4-forum-example> php vendor/bin/phpunit
PHPUnit 9.6.12 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.5 with Xdebug 3.1.4
Configuration: P:\MyGitHubWork\codeigniter4-forum-example\phpunit.xml.dist
Random Seed:   1694842617

.....................................                             37 / 37 (100%)

Nexus\PHPUnit\Extension\Tachycardia identified this sole slow test:
+-------------------------------------------------------------------------+---------------+-------------+
| Test Case                                                               | Time Consumed | Time Limit  |
+-------------------------------------------------------------------------+---------------+-------------+
| Tests\\Controllers\\ThreadControllerTest::testCanUserUpdateHisOwnThread | 00:00:01.01   | 00:00:01.00 |
+-------------------------------------------------------------------------+---------------+-------------+


Time: 00:27.288, Memory: 56.00 MB

OK (37 tests, 140 assertions)

Generating code coverage report in PHP format ... done [00:00.002]

Generating code coverage report in Clover XML format ... done [00:00.125]

Generating code coverage report in HTML format ... done [00:00.515]
```
Why I could not reproduce the error locally?!

michalsn:
The error was popping out randomly - depending on what order tests were executed. Thanks for testing, though.

Anyway... not sure if this is the best way to do this, but now it should be fixed.

datamweb:
@michalsn "Create a new reply" not work for me!
```console
htmx.min.js:1 htmx:targetError
```

michalsn:
@datamweb Thank you! Should be fixed now.

I forgot I was changing the way replies are presented.

datamweb (approved):
Thanks! excellent.

What I noticed now is that we don't have an cancel button when editing.

michalsn:
That's true. We can add this in the future.

lonnieezell (approved):

PR #59, opened 2023-09-18
feat: tags

### LINKED ISSUE #12: Users should be able to use tags within discussions.
As a user I should be able to use tags within my posts and have them be recognized as such. These are used to help provide finer-grained self-categorization of posts throughout the system. Eventually, a user would be able to search and see hashtag results. 

- This should be presented as a separate input field where the user can enter tags when creating/editing a post
-  The tags should be single-words only (can use dashes and underscores)
- Should convert to a visual representation that it's a tag when they hit space, tab, or enter.

michalsn:
This PR adds tag support for the thread.

The tags are validated in the frontend and backend. 

* I assumed we would use tags only for threads, so the table design is simplified
* When we click a tag, we can browse all threads tagged with that tag
* Only dashes are allowed to separate words. Allowing underscores would potentially create duplicates of the same tags
* The maximum length of the single tag is 20 characters
* The maximum number of tags per thread is 5

Closes #12

### DIFF
--- a/app/Controllers/Discussions/TagController.php
+++ b/app/Controllers/Discussions/TagController.php
@@ -33,7 +33,7 @@ class TagController extends BaseController
         $rules = [
             'perPage'     => ['in_list[20]'],
             'search.type' => ['permit_empty', 'in_list[' . implode(',', array_keys($types)) . ']'],
-            'search.tag'  => ['required', 'regex_match[/^[a-z0-9-]{0,20}$/]']
+            'search.tag'  => ['required', 'regex_match[/^[a-z0-9-]{0,20}$/]'],
         ];
 
         if (! $this->validateData($table, $rules)) {
--- a/app/Controllers/Discussions/TagController.php
+++ b/app/Controllers/Discussions/TagController.php
@@ -3,25 +3,25 @@
 namespace App\Controllers\Discussions;
 
 use App\Controllers\BaseController;
-use App\Models\PostModel;
 use App\Models\ThreadModel;
-use CodeIgniter\Exceptions\PageNotFoundException;
 use InvalidArgumentException;
 
 /**
  * Class Home
  */
-class DiscussionController extends BaseController
+class TagController extends BaseController
 {
     /**
      * Display a standard forum-style list of discussions.
      */
-    public function list(): string
+    public function get(string $tagSlug)
     {
         $table = [
             'perPage' => $this->request->getGet('perPage') ?? 20,
             'search'  => $this->request->getGet('search') ?? [],
         ];
+
+        $table['search']['tag'] = $tagSlug;
 
         $types = [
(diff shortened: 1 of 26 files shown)


app/Controllers/Discussions/TagController.php
```diff
@@ -0,0 +1,62 @@
...
+ * Class Home
+ */
+class TagController extends BaseController
+{
+    /**
+     * Display a standard forum-style list of discussions.
+     */
+    public function get(string $tagSlug)
+    {
+        $table = [
```
lonnieezell, line 19: At some point, I'd love to refactor all of our discussion searching into a `search` method on the model to clean it up a little, but that's just me thinking out loud.

app/Models/TagModel.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * Find or create tag and return tag ID.
+     *
+     * @throws ReflectionException
+     */
+    public function findOrCreateId(string $name): int
+    {
+        $tagId = $this->where('name', $name)->first()?->id;
```
lonnieezell, line 61: Should we standardize all tags to lowercase? thinking maybe we should since the name is used in the URI.

lonnieezell (commented):
Looks great! I just had the one question about normalizing tag names to lowercase I'd like to get  your thoughts on.

app/Controllers/Discussions/TagController.php
michalsn: On the model level, we're already using only one method: `ThreadModel::forList`.

Do you mean searching for threads, in the top menu bar? Or one method in the controller to display threads list, tags, and categories?

app/Models/TagModel.php
michalsn: Sorry, I forgot to mention it. We already allow only lowercase letters for the tag name. It's validated in the backend too, so I would say we're safe here.

michalsn: Unless we don't want to be so restrictive and allow uppercase and then convert everything to lowercase...

lonnieezell: Ok, I missed that! I think that's fine for now. We can always update later if it gets to feeling restrictive.

app/Controllers/Discussions/TagController.php
lonnieezell: I just meant handling all of the filtering, sorting, etc. It feels like we're using similar code in several places, but maybe they're all different models and wasn't thinking straight last night :)

lonnieezell (approved):

app/Controllers/Discussions/TagController.php
michalsn: I made some refactoring today, and I see redundant code in the DiscussionController... however, for now, I think I will leave it as it is until we are sure the functionality is finished.

PR #62, opened 2023-09-20
Refactoring and updates

michalsn:
This PR makes some refactoring and updates:

* fixes for top nav (wasn't properly aligned with daisyUI v3)
* fixes for categories nav (wasn't displaying correctly)
* adds discussions by category
* moves discussions by tag to DiscussionController
* adds cancel button when creating/editing thread/post
* and finally adds some more tests

Sorry for mixing so many things in one PR. I will try to avoid this in the future.


lonnieezell (approved):
Looks like all good changes. Thanks!

PR #66, opened 2023-09-21
feat: display avatar

michalsn:
This PR makes the avatar display next to the author of the post.

### DIFF
--- a/app/Concerns/HasAuthorsAndEditors.php
+++ b/app/Concerns/HasAuthorsAndEditors.php
@@ -4,7 +4,6 @@ namespace App\Concerns;
 
 use App\Models\UserModel;
 use CodeIgniter\Entity\Entity;
-use CodeIgniter\Shield\Authentication\Authenticators\Session;
 
 trait HasAuthorsAndEditors
 {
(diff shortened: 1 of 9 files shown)


app/Concerns/HasAuthorsAndEditors.php
```diff
@@ -41,6 +42,16 @@ public function withUsers(array|Entity|null $records): array|Entity|null
         array_walk($records, static function (&$post) use ($users) {
             $post->author = $users[$post->author_id] ?? null;
             $post->editor = $users[$post->editor_id] ?? null;
+
+            // Get email to grab avatar later
+            if ($identityAuthor = $post->author?->getIdentities(Session::ID_TYPE_EMAIL_PASSWORD)[0] ?? null) {
```
lonnieezell: I haven't followed recent updates on Shield, but you should be able to just pull `author?->email` and have Shield do the logic in the background, right?

app/Config/Users.php
```diff
@@ -24,7 +24,7 @@ class Users extends BaseConfig
      * The default image type to use from Gravatar if they do not have an
      * image for that user.
      */
-    public $gravatarDefault = 'blank';
+    public $gravatarDefault = 'mp';
```
lonnieezell: Doesn't this cover up their initials when they don't have an avatar?

app/Concerns/HasAuthorsAndEditors.php
michalsn: You're right. I haven't really used Shield before and it seems I missed the basic functionality. Fixed.

app/Config/Users.php
michalsn: I missed the `renderAvatar` method and related view. Now I understand why we were using the `blank` option. 

However, we won't display initials since the gravatar always generates a link when gravatar is enabled: https://github.com/lonnieezell/forum-example/blob/develop/app/Views/users/_avatar.php#L4

lonnieezell (approved):
Looks good. Thanks!

PR #67, opened 2023-09-22
feat: image upload

michalsn:
This PR adds support for attaching images to the threads and posts.

I resigned from other types of files since no one reasonably would download a ZIP file from a stranger. Larger projects can be hosted on GitHub or Gist.

I took the approach of making all the files publicly available. Therefore there is no option to restrict access to the files. I find it reasonable since no one should send any sensitive data, and requiring to log in to see the images was never a standard in our forum.

More info can be found in `images.md` file.

### DIFF
--- a/app/Concerns/HasTags.php
+++ b/app/Concerns/HasTags.php
@@ -33,7 +33,7 @@ trait HasTags
      */
     public function addTags(array $tags): static
     {
-        $this->tags = array_unique(array_merge($this->tags, $tags));
+        $this->tags = array_filter(array_unique(array_merge($this->tags, $tags)));
 
         $this->withTags();
 
@@ -45,7 +45,7 @@ trait HasTags
      */
     public function removeTags(array $tags): static
     {
-        $this->tags = array_diff($this->tags, $tags);
+        $this->tags = array_filter(array_diff($this->tags, $tags));
 
         $this->withTags();
 
@@ -57,7 +57,7 @@ trait HasTags
      */
     public function replaceTags(array $tags): static
     {
-        $this->tags = array_unique($tags);
+        $this->tags = array_filter(array_unique($tags));
 
         $this->withTags();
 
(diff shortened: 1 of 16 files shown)


app/Concerns/HasTags.php
```diff
@@ -33,7 +33,7 @@ public function withTags(): static
      */
     public function addTags(array $tags): static
     {
-        $this->tags = array_unique(array_merge($this->tags, $tags));
+        $this->tags = array_filter(array_unique(array_merge($this->tags, $tags)));
```
lonnieezell, line 36: Would a collection class help here? I've got one that I built a bit ago but have never used. I just made it public if you want to check it out: [myth-collections](https://github.com/lonnieezell/myth-collection)

app/Controllers/Discussions/ImageController.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        if (! $this->policy->can('images.upload')) {
+            return $this->response->setStatusCode(401)->setJSON([
+                'error' => 'You are not allowed to upload images.',
+            ]);
+        }
+
+        if (! $this->validate([
+            'image' => [
+                'uploaded[image]', 'max_size[image,2048]',
```
lonnieezell, line 29: I think we are going to want the image size to be configurable at some point down the road, but we can worry about that later.

lonnieezell (approved):

app/Concerns/HasTags.php
michalsn: This library looks pretty cool, although I don't see an equivalent to the `array_unique` function (I could be wrong). The `array_filter` function is used here to remove empty values. It would be cool if myth-collections could make it happen without the need to write custom implementation in the `filter` method (but again, I could be wrong).

Right now I think we would produce much more code to achieve the same effect.

app/Controllers/Discussions/ImageController.php
michalsn: Will do!

app/Concerns/HasTags.php
lonnieezell: You're probably right. And I am surprised I never put in a filter function. I'll have to remedy that at some point. It's too useful.

PR #68, opened 2023-09-23
Use the CI tasks library for our cron job.

lonnieezell:
We have an official [cron/tasks library](https://github.com/codeigniter4/tasks), but I really need to get in there and finish the docs so that we can do a release. It kind of got dropped by the wayside a long time ago. Incorporating it in this, though, as it makes sense. And will give the impetus needed to get a release of it out. 

Scheduled the `cleanup:images` command every hour.

### DIFF
--- a/app/Config/Tasks.php
+++ b/app/Config/Tasks.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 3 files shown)


michalsn (approved):
I have never used the task library, but it looks great.

PR #71, opened 2023-09-24
Initial page wrapper for the member Account page.

lonnieezell:
- new Account page with: 
  -  header with basic info about user
  - sidebar with links to expected base pages

Includes:
- new fields for the user profile
- fixes for avatars to actually display the user's initials.


michalsn (approved):
Looks great!

CS fixes can be done later.

PR #79, opened 2023-09-28
Disable callbacks for Concerns

michalsn:
This PR disables callbacks in all `Concerns` and methods used exclusively as`callbacks`.

If the method is used as a callback, then triggering any other callback when running the callback is unnecessary. It might lead to a situation where we will be making unnecessary queries.

For that reason, we should disable callbacks in these types of methods unless using the callback actually brings some value.


lonnieezell (approved):
I had forgotten this method even exists, but makes perfect sense here. Thanks.

PR #80, opened 2023-09-29
Switch to tags library

michalsn:
This PR switches the tag handling to an [external library](https://github.com/michalsn/codeigniter-tags). (it uses `myth-collection` for tags).

The code is improved and better tested - this also means less code to maintain here.

The database tables have changed, but I didn't add table migration for deleting the old tables, so we have to start over with migration and seeding to make it work properly.


lonnieezell (approved):
Sure - try to bribe me by using my own library against me lol 

Looks good to me!

michalsn:
Haha, the bribe was intended 😅

lonnieezell:
It does explain why you were adding so many PRs. :)

PR #86, opened 2023-10-05
Make Account pages mobile-responsive.

lonnieezell:
Unfortunately `$this->renderSection()` doesn't work 2 levels deep, so had to use a Cell, but it works.


michalsn (approved):
Nav Cell is a nice idea, though we should ensure it's called with the `hx-boost` when in the sidebar.

lonnieezell:
Got to thinking - the Cell was actually more complex than needed. Simplified it to just a simple view. 

Also fixed the avatar display.

PR #91, opened 2023-10-06
feat: notifications

michalsn:
This PR adds some notification options about new posts.

I thought it would be useful to make it as simple as possible, that is, only the main settings for notifications - without signing up for them when creating a post/thread.

What do you guys think about the current types of notifications? Is it too much, or is something missing?

To do:
* add a link to open the discussion/post from email in the browser
* provide a link to opt out of notifications for a given thread in email
* add a queue system for sending emails (probably based on the DB since this is our only available option?)

But what I have listed above, I will add in separate PRs in the future.

### DIFF
--- a/app/Events/NewPostEvent.php
+++ b/app/Events/NewPostEvent.php
@@ -10,8 +10,21 @@ use App\Models\PostModel;
 
 class NewPostEvent
 {
+    /**
+     * Number of notifications sent.
+     */
+    private int $count = 0;
+
     public function __construct(protected Thread $thread, protected Post $post)
     {
+    }
+
+    /**
+     * Get number of notifications sent.
+     */
+    public function getCount(): int
+    {
+        return $this->count;
     }
 
     /**
@@ -60,6 +73,8 @@ class NewPostEvent
             return false;
         }
 
+        $this->count++;
+
         // send notification
         return $this->sendNotification($this->thread->author, $this->thread, $this->post);
     }
@@ -100,6 +115,7 @@ class NewPostEvent
             if ($setting->email_post === true) {
                 // send notification
                 $this->sendNotification($this->post->author, $this->thread, $this->post);
+                $this->count++;
 
                 continue;
             }
@@ -110,6 +126,7 @@ class NewPostEvent
                 if ($postReplyTo?->author_id === $setting->user_id) {
                     // send notification
                     $this->sendNotification($this->post->author, $this->thread, $this->post);
+                    $this->count++;
 
                     continue;
                 }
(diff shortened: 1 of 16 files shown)


app/Events/NewPostEvent.php
```diff
@@ -0,0 +1,145 @@
...
+     * Send email notification.
+     *
+     * @todo Make it run in the background (queueNotification).
+     */
+    private function sendNotification(User $recipient, Thread $thread, Post $post): bool
+    {
+        helper('text');
+
+        return service('email', false)
+            ->setFrom('notifications@myth.forum', config('App')->siteName)
```
lonnieezell, line 54: The email address should be a config setting also, right?

app/Events/NewPostEvent.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Handle post related notifications.
+     */
+    private function handlePostNotifications(): bool
+    {
+        $postModel = model(PostModel::class);
+
+        // get all post authors except the current on
```
lonnieezell:
```suggestion
        // get all post authors except the current one
```

lonnieezell (changes requested):
In general I think it looks great. A couple of small suggestions. 

At first I wondered if we should use MGatner's user settings library, but I just looked and I guess he removed all of his CodeIgntier libraries. Guess he didn't want to support them anymore. Bummer. In that case I have an `alerts` package in Bonfire I need to do something about.

lonnieezell:
I think those three settings are perfect. And I always found it annoying that MyBB asked you if you wanted to subscribe on every thread. Should be a global thing like you have it here. 

As for queue - I agree that's the way to go, and I'd keep that simple and db driven also. There's been a couple stabs at implementing a queue for CI in the past, but I've never had the time to dive into their implementations and clean things up and take it across the finish line.

app/Events/NewPostEvent.php
michalsn: Yes, I forgot about that. I think we can use `config('Email')->fromEmail`, or just skip `setFrom()` altogether and fill in the `$fromEmail` and `$fromName` config variables - then the Email library should use those variables by default.

michalsn:
MGatner had his libraries in a separate organization and it looks like everything is in place: https://github.com/tattersoftware

Thanks for the feedback! We'll see how it goes, but it's good to know we're on the right track with notifications.

lonnieezell:
> MGatner had his libraries in a separate organization and it looks like everything is in place: https://github.com/tattersoftware
> 
> Thanks for the feedback! We'll see how it goes, but it's good to know we're on the right track with notifications.

Doh! You're right. It's been awhile and I'm worn out today lol.

lonnieezell:
> Thanks for the feedback! We'll see how it goes, but it's good to know we're on the right track with notifications.

One thing I think we might consider adding down the road is a "Mute" button on a thread incase we're tired of hearing about a particular topic.

lonnieezell (approved):

michalsn:
> One thing I think we might consider adding down the road is a "Mute" button on a thread incase we're tired of hearing about a particular topic.

Good point. Adding this feature at the page level, not just via email, should not be a problem.

PR #94, opened 2023-10-11
feat: mute thread notifications

michalsn:
This PR adds an option to mute notifications for specified discussions.

When we get an email notification about a new post, we have an option to mute further notifications from that discussion.

The link does not require logging into the site since it's signed, so there is no way to modify it. I smuggled my library here: https://github.com/michalsn/codeigniter-signed-url

Also at the page level, on the discussion page, there is a button that allows us to do the same. It is only visible when we are logged in and have notifications enabled in the settings.

From other things, I also removed an unused cell.

Last but not least... a question about our icons. We use heroicons, right? I added one new icon and took it from here: https://heroicons.dev is it the correct place?


lonnieezell (approved):
Looks good.

lonnieezell:
Yes. @michalsn you are correct - heroicons it is.

lonnieezell:
Though I've always grabbed them from: https://heroicons.com/

michalsn:
Thanks! I asked because my SVG code was a bit different from other icons. Now I know the correct place to grab them.

PR #98, opened 2023-10-15
feat: add email queue

michalsn:
This PR adds an email queue for sending new post notifications.

I decided to make a separate project to support queues: https://github.com/michalsn/codeigniter-queue, so that we could focus just on writing the actual code that should be run.

The queue worker for emails is started via the Tasks library. It will try to send 20 emails and then will quit. If there are no emails to send it will quit immediately. We can adjust this however we want, but this looks like a good starting point.

For other things, the queue doesn't support objects, so they have to implement a `jsonSerialize` interface. I didn't want to use the `serialize`/`unserialize` due to security issues. Not an ideal solution, but it works for now.


datamweb (approved):
I like this procedure of separating parts into different packages.
It looks good.

jozefrebjak:
@michalsn it looks great. I'm trying implement queue in my app and I follow this PR, but I have a problem where is not possible to use `setting` helper in `process()` method. I believe the same problem will be there, because email config should be set via UI and then it's saved in DB. 

Even if I add `helper('setting');` in `process()` method I'm still getting failed job with message `Exception: 0 - Call to undefined function IctSolutions\Admin\Jobs\setting()
file: /app/modules/Admin/Jobs/EmailActivityMessageNotification.php:16`

Now is possible to set email config only via `.env` I think and directly via `app/Config/Email`.

michalsn:
@jozefrebjak Sorry, but I cannot reproduce the error.

Before sending a job to the queue:
```php
helper('setting');
setting('App.baseURL', 'https://example123.com');

service('queue')->push('emails', 'email', ['message' => 'Sample message']);
```
The job:
```php
class Email extends BaseJob implements JobInterface
{
    public function process()
    {
        helper('setting');
        $name = setting('App.baseURL');
        file_put_contents(WRITEPATH . 'baseurl.txt', $name);
     }
}
```
And it's working just fine.

```cli
php spark queue:work emails

Listening for the jobs with the queue: emails

Starting a new job: email, with ID: 11
The processing of this job was successful
It took: 0.0035 sec

No job in the queue. Waiting...

```

jozefrebjak:
@michalsn thank you for your message. Yes, you are right, I needed to stop queue and start again `php spark queue:work emails -max 20`. I added 

```php
        helper('setting');

        $config = [
            'fromEmail'     => setting('Email.fromEmail'),
            'fromName'      => setting('Email.fromName'),
            'userAgent'     => setting('Email.userAgent'),
            'protocol'      => setting('Email.protocol'),
            'mailPath'      => setting('Email.mailPath'),
            'SMTPHost'      => setting('Email.SMTPHost'),
            'SMTPUser'      => setting('Email.SMTPUser'),
            'SMTPPass'      => setting('Email.SMTPPass'),
            'SMTPPort'      => setting('Email.SMTPPort'),
            'SMTPTimeout'   => setting('Email.SMTPTimeout'),
            'SMTPKeepAlive' => setting('Email.SMTPKeepAlive'),
            'SMTPCrypto'    => setting('Email.SMTPCrypto'),
            'wordWrap'      => setting('Email.wordWrap'),
            'wrapChars'     => setting('Email.wrapChars'),
            'mailType'      => setting('Email.mailType'),
            'charset'       => setting('Email.charset'),
            'validate'      => setting('Email.validate'),
            'priority'      => setting('Email.priority'),
            'CRLF'          => setting('Email.CRLF'),
            'newline'       => setting('Email.newline'),
            'BCCBatchMode'  => setting('Email.BCCBatchMode'),
            'BCCBatchSize'  => setting('Email.BCCBatchSize'),
            'DSN'           => setting('Email.DSN'),
        ];
```

and 

```php
 $email ->initialize($config);
```

to make it work. Another option is to use `emailer` helper form shield.

michalsn:
@jozefrebjak I'm glad it worked for you.

lonnieezell (approved):
Looks great! And I think having the queue as a separate package is a great idea.

PR #100, opened 2023-10-18
feat: Basic security page

lonnieezell:
- Displays the basic security page to the account holder. 
- Displays information about the last 5 logins
- Provides a way to reset the password
- Provides a way to logout of all remembered devices. 
- Allows deleting their account - this fires an event so that other actions can be taken, like extra GDPR data removal, etc.

A few things I think we need to do in the future but is not included with this are: 
- a helper function for displaying dates based on the user's timezone
- loading states when triggering AJAX requests. 
- a way to show alerts/toasts globally?


app/Controllers/Account/SecurityController.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace App\Controllers\Account;
+
+use App\Controllers\BaseController;
+use CodeIgniter\HTTP\RedirectResponse;
+use CodeIgniter\HTTP\Exceptions\HTTPException;
+use CodeIgniter\Shield\Authentication\AuthenticationException;
+use CodeIgniter\Shield\Models\RememberModel;
+use ReflectionException;
```
datamweb, line 10: use App\Entities\User;

datamweb:
```suggestion
use ReflectionException;
use App\Entities\User;
```

app/Controllers/Account/SecurityController.php
```diff
@@ -0,0 +1,44 @@
...
+class SecurityController extends BaseController
+{
+    /**
+     * Manage the user's account security.
+     */
+    public function index()
+    {
+        return $this->render('account/security/security', [
+            'user' => auth()->user(),
+            'logins' => auth()->user()->logins(7),
```
datamweb:
```suggestion
        /** @var User::class $user */
        $user = auth()->user();
        
        return $this->render('account/security/security', [
            'user' => $user,
            'logins' => $user->logins(7),
```

app/Controllers/Account/SecurityController.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Logs the user out of all sessions, by deleting all remember-me tokens.
+     * Logs the user out and redirects the user to the login page.
+     */
+    public function logoutAll()
+    {
+        try {
+            auth()->forget(auth()->user());
```
datamweb, line 41: I don't know exactly what you mean, but if you want to remove remember-me user, I think the following code snippet is good.
```php
use CodeIgniter\Shield\Models\RememberModel;
// ...

$rememberModel = model(RememberModel::class);

/** @var User::class $user */
$user = auth()->user();

$rememberModel->purgeRememberTokens($user);
```

lonnieezell: That's basically what the `forget()` method on the Session Authenticator does. 

https://github.com/codeigniter4/shield/blob/50fb634742c2b4379c51855ae7175d77c34eeac3/src/Authentication/Authenticators/Session.php#L854

datamweb: Thanks, I hadn't seen this one.

datamweb:
Please rebase.
Please follow command  `php vendor/bin/rector --dry-run` to fix the Rector error.

Regarding error `Call to an undefined method CodeIgniter\Shield\Auth::forget()` in PR https://github.com/codeigniter4/shield/pull/923, it was resolved. 

~But I don't know what is the reason for:~
```console
 ------ ---------------------------------------------------------------- 
  Line   app/Controllers/Account/SecurityController.php                  
 ------ ---------------------------------------------------------------- 
  22     PHPDoc tag @var for variable $user contains unresolvable type.  
 ------ ---------------------------------------------------------------- 
```
fix by:
```diff 
-/** @var User::class $user */
-$user = auth()->user();
+/** @var User $user */
+$user = auth()->user();
```
```console
PS P:\MyGitHubWork\codeigniter4-forum-example> php vendor/bin/phpstan
Note: Using configuration file P:\MyGitHubWork\codeigniter4-forum-example\phpstan.neon.dist.
 76/76 [============================] 100%


                                                                                                                        
 [OK] No errors                                                                                                         
                                                                                                                        
```

app/Views/account/security/security.php
```diff
@@ -0,0 +1,49 @@
...
+            <!-- Password reset button and description -->
+            <div>
+                <h3 class="font-bold">Log Out of All Devices</h3>
+                <p class="text-sm opacity-50">
+                    Log out of all devices, including this one.
+                </p>
+            </div>
+            <div class="flex align-middle h-full">
+                <button class="btn btn-outline btn-warning" <?= !$isRemembered ? 'disabled' : '' ?>
+                    hx-post="<?= url_to('account-security-logout-all') ?>"
```
michalsn, line 31:
```suggestion
                    hx-post="<?= url_to('account-security-logout-all') ?>"
                    hx-headers='<?= json_encode([csrf_header() => csrf_hash()]) ?>'
```

Normally we use form tags, but here we need to add CSRF manually.

michalsn (approved):
Looks good to me.

One thing I would like to consider (for later) is not immediately deleting the account. The current way is safe because we have to provide the current password, but an even better option would be to schedule the user's deletion for later, such as in 7 days (or similar). Which gives him or her time to think about it, and the option of an email notification that the process has been started for added security. But as I said, this would be something for later, as something we can add to the todo.

lonnieezell:
@michalsn I think that's a great idea. Created a new task for it.

lonnieezell:
Ah, yes, now we're getting into the static analysis gripes I love so much. Just sucks the joy out of coding....

michalsn:
Yeah... I would just merge it. We can deal with that later.

PR #102, opened 2023-10-20
Enable CSRF

michalsn:
This PR enables the CSRF filter and upgrades CodeIgniter to version 4.4.2.

* v4.4.2 was required to enable CSRF in our case (ref: https://github.com/codeigniter4/CodeIgniter4/pull/7915)
* I also updated the readme to include links to the icons and markdown editor

resolved #103

### DIFF
--- a/tests/Controllers/PostControllerTest.php
+++ b/tests/Controllers/PostControllerTest.php
@@ -52,11 +52,13 @@ final class PostControllerTest extends TestCase
         ]);
 
         $fileUrl  = base_url('uploads/' . $user->id . '/test_image1.jpg');
-        $response = $this->actingAs($user)->post('posts/1', [
-            'thread_id' => '1',
-            'reply_to'  => '',
-            'body'      => 'Sample body for post ![](' . $fileUrl . ')',
-        ]);
+        $response = $this
+            ->withHeaders([csrf_header() => csrf_hash()])
+            ->actingAs($user)->post('posts/1', [
+                'thread_id' => '1',
+                'reply_to'  => '',
+                'body'      => 'Sample body for post ![](' . $fileUrl . ')',
+            ]);
 
         $response->assertStatus(200);
         $response->seeElement('.post-meta');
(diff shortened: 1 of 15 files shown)


tests/Controllers/PostControllerTest.php
```diff
@@ -52,11 +52,13 @@ public function testCanUserCreateAPost()
         ]);
 
         $fileUrl  = base_url('uploads/' . $user->id . '/test_image1.jpg');
-        $response = $this->actingAs($user)->post('posts/1', [
-            'thread_id' => '1',
-            'reply_to'  => '',
-            'body'      => 'Sample body for post ![](' . $fileUrl . ')',
-        ]);
+        $response = $this
+            ->withHeaders([csrf_header() => csrf_hash()])
```
lonnieezell, line 56: We should consider an addition either here or in CI itself that simplifies this, like `->withCSRF(()` at some point...

lonnieezell: I take that back. This isn't something the devs should have to think about at all, which means it's definitely something for the core, I think.

lonnieezell (approved):

tests/Controllers/PostControllerTest.php
michalsn: To be honest, I was a little surprised that I had to add it because it seemed to me that in the test environment, CSRF checking was always disabled, but maybe it was in CI3.... or I'm wrong in general 😅

datamweb (approved):

tests/Controllers/PostControllerTest.php
lonnieezell: Yeah, not sure, honestly. Will be a bit before I can dig into the core code to see what's going on. Trying to wrap up the security page tonight. I probably should have broken that up into a few different PRs, but too late now :) 

To be fair, I don't recall ever having to do that manually in the past. If we do, it's definitely a step backwards and potential blocker to getting our community more into testing.

PR #104, opened 2023-10-21
fix: Attempt to read property "email_thread" on null 

datamweb:
deleted all rows from `notification_settings` see http://localhost:8080/account/notifications


michalsn:
There is no option that the user has no notification entry: https://github.com/lonnieezell/forum-example/blob/develop/app/Models/UserModel.php#L26

Your database data were probably created earlier than this code. Please, seed your database one more time.
We are at such an early stage that I do not want to create code that will not be required.

datamweb:
@michalsn You are right. Thank you!

michalsn:
@datamweb Also thank you for trying to fix it.

PR #105, opened 2023-10-22
Display my posts in `account/posts`

michalsn:
Display "My posts" that link directly to the thread.

### DIFF
--- a/app/Entities/Post.php
+++ b/app/Entities/Post.php
@@ -3,6 +3,8 @@
 namespace App\Entities;
 
 use App\Concerns\RendersContent;
+use App\Models\CategoryModel;
+use App\Models\ThreadModel;
 use CodeIgniter\Entity\Entity;
 
 class Post extends Entity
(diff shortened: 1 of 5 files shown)


app/Entities/Post.php
```diff
@@ -27,13 +27,21 @@ class Post extends Entity
...
-        $categorySlug = $category === null ?
-            model('CategoryModel')->find($this->category_id)->slug :
-            $category->slug;
-
-        $threadSlug = $thread === null ?
-            model('ThreadModel')->find($this->thread_id)->slug :
-            $thread->slug;
+        if (! isset($this->category_slug)) {
+            $categorySlug = $category === null ?
+                model('CategoryModel')->find($this->category_id)->slug :
```
datamweb: Let's define a standard from now on model(CategoryModel::class) or model('CategoryModel')?
We are using arbitrary in different sections, this destroys code integrity.

michalsn: Good point. Personally, I like `model(CategoryModel::class)` more and that is a convention mostly used in the codebase I think.

datamweb (approved):
Screenshot 2023-10-22 140259

LGTM!
Thanks.

lonnieezell (approved):

PR #117, opened 2023-10-27
feat: Timed user deletion

michalsn:
This PR adds timed user deletion.

* When the user is soft-deleted, we also soft-delete all posts, ~replies to posts~, and discussions.
* We are recalculating all the stats related to soft-deleted users and all other users who were impacted (impacted by removing their posts or replies to posts).
* We send an email message with the information that this account has been scheduled for deletion. In every email, we also provide a special link to restore the account.
* Account is scheduled to be removed after 7 days. This is configurable via the `Config\Forum::accountDeleteAfter` variable
* 3 and 1 day before scheduled deletion we are sending a reminder with the information and a link to restore the account.
* When we finally remove the account, we also remove all the images that are not needed anymore

There is one thing that bugs me... it's recalculating stats when we soft-delete a user. Do you guys think the path I took is okay? I'm not sure.

Also, right now we depend on the `deleted_at` field value when we're restoring the user and all related data. Not an ideal solution but I hated the idea of adding another column to 3 tables just for this.


lonnieezell:
I've gone back and forth, but I think updating the stats upon request is fine. It will match with what will be shown to other visitors on the site.

_docs/users.md
```diff
@@ -0,0 +1,28 @@
+# Users
+
+## Removing users
+
+Any user can request the deletion of their account. This can be done via user settings.
+
+### How is this working
+
+When a user deletes their account, it is soft-deleted. Along with it, all user's posts, 
+replies to her/his posts and created discussions are deleted in the same way.
```
lonnieezell: I think for replies to their posts, we should just change the parent of the next post? That allows discussions to remain intact, even though the single post was deleted. I've seen that happen on several other systems. What do you think? 

I think it makes sense to delete all posts for their discussions, of course.

michalsn: Well... this is tough. I took the current approach because there are many disadvantages when we leave replies to deleted posts.

* If we change the parent, the replies may be completely out of context of the other conversation's point of view
* Replies may interfere with the ongoing conversation (different publication times).
* What if there are no other parent posts? Should we make all replies top-level posts? We will face the same problems as above.
* There will be no way to undo this action when the user decides to recover their account.

IMO the ideal solution, when deleting a single post that has replies, would be:
* Not display the username and message of that post, but leave the replies so that the conversation structure can remain intact.
* This would make recovery easier, but would be difficult to handle during hard deletion... we would need an additional variable to indicate this special type of post (deleted, but not for real).

michalsn:
> I've gone back and forth, but I think updating the stats upon request is fine. It will match with what will be shown to other visitors on the site.

Yes, the other alternatives I see, are:
* don't touch the stats at all
* update them regularly once a day for all the data

I like the idea of updating instantly, but it's not an easy one.

_docs/users.md
lonnieezell: I think there are relatively simple solutions to the issues you had, BUT - I rather like your ideal solution, honestly. I'm all for that if you want to do it.

app/Entities/UserDelete.php
```diff
@@ -0,0 +1,26 @@
...
+    protected $datamap = [];
+    protected $dates   = ['scheduled_at', 'created_at'];
+    protected $casts   = [
+        'user_id' => 'integer',
+    ];
+
+    /**
+     * Generate link to restore user account.
+     */
+    public function restore(): string
```
lonnieezell:
```suggestion
    public function restoreLink(): string
```

My initial reaction with the function name was that this handled restoring the user, but since it just generates the link, can we name it a little more specifically?

app/Views/_emails/account_delete_reminder.php
```diff
@@ -0,0 +1,21 @@
...
+
+We're reaching out to let you know that your account is scheduled to be permanently removed on <strong><?= $userDelete->scheduled_at->toDateString(); ?></strong>.
+
+<br><br>
+
+If you haven't changed your mind on this, then you don't have to make any actions.
+
+<br><br>
+
+But if you have reconsidered your desicion, you can <a href="<?= $userDelete->restore() ?>">click here</a> to restore your account.
```
lonnieezell:
```suggestion
But if you have reconsidered your decision, you can <a href="<?= $userDelete->restore() ?>">click here</a> to restore your account.
```

lonnieezell:
I think it all looks great once we finalize how we're handling  deleted posts.

_docs/users.md
michalsn: Ok, I will try to implement this.

app/Entities/UserDelete.php
michalsn: You're right, thanks!

michalsn:
Ok, done.

lonnieezell (approved):
Looks great. Thanks!

PR #118, opened 2023-10-30
feat: Alerts

michalsn:
This PR adds an Alerts library, as discussed in: https://github.com/lonnieezell/forum-example/discussions/108

Also, there are small changes in the notification settings.

### DIFF
--- a/_docs/alerts.md
+++ b/_docs/alerts.md
@@ -1,0 +1,69 @@
+# Alerts
+
+The basic `Alerts` library allows us to display notifications about the results of actions performed by the user.
+
+Since we use [htmx](https://htmx.org) in most places, we need to have a system that allows us to display inline 
+notifications. But also be able to display the notification after redirect.
+
+That's why this custom alerts library was created.
+
+## How is this working
+
+At the moment, alerts are built on top of the [Toast](https://daisyui.com/components/toast/) component.
+
+Alert types are set based on the available classes in the Toast component, but we should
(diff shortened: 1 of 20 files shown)


_docs/alerts.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Changing position of alerts
+
+By default, alerts are displayed in the right bottom of the screen.
+To change that we have to edit view file: `app/Views/_alerts/container`.
+
+### Auto-hiding alerts
+
+By default, each alert is hidden after 5 seconds. We can change that by
+changing config variable: `Config\Forum::alertDisplayTime`.
```
datamweb, line 69:
```suggestion
changing config variable: `Config\Forum::alertDisplayTime`.
```
Honestly, I believe, this value should be dynamically changeable. Because the length of the message text may be long in some cases, so the programmer should be able to set this amount of time for the long text.

e.g
 
```php
alerts()->set('success', 'Message success')->hidingAfter(3*1000);
```

lonnieezell: I don't remember where I saw it but somewhere recently they had a great solution: It had a thin "progress bar" on the bottom of the toast that would show how long it was lasting. When you hovered over it, or took the focus off of the page, the countdown would stop until you either came back to the page or the cursor moved out of it.

lonnieezell: I do think this is something that can be iterated over and shouldn't hold up this PR, though.

lonnieezell (approved):

PR #123, opened 2023-11-03
feat: View and edit basic profile fields

lonnieezell:
Adds basic profile editing to the account area. 

A few things missing that will be addressed in later PRs: 
- needs ability to edit avatar
- the signature needs a preview
- should be able to set the timezone. I would like to look at making a more user-friendly dropdown solution for this, though, than the stock CI one as that long list of timezones is overwhelming. Probably filter it down based off of country when possible, and then display it timezone names/offset, like "-5 : CST", or something.


michalsn (approved):
Looks good. Nice plan for future improvements.

~When it comes to time zones, I'm more of a UTC guy. Though, I'm not sure what would be more universal/common.~
Sorry... I guess it's too late for rational thinking for me 😅

lonnieezell:
Honestly, I'm not even sure if we need to implement timezone support here yet. I think the vast majority of places we show dates/times so far have been humanized...

PR #125, opened 2023-11-04
Alerts update

michalsn:
This PR addresses requests mentioned in the [#118](https://github.com/lonnieezell/forum-example/pull/118#discussion_r1377244694)

* configurable display time for alert via 3rd (optional) parameter 
```php
alerts()->set('success', 'alert message', 10);
```
* pause alert's display time on hover 
* show a progress bar for every alert (this doesn't look very good right now because I use a standard progress component, but it can be fixed in the future)

### DIFF
--- a/app/Filters/AlertsFilter.php
+++ b/app/Filters/AlertsFilter.php
@@ -26,6 +26,8 @@ class AlertsFilter implements FilterInterface
      */
     public function after(RequestInterface $request, ResponseInterface $response, $arguments = null)
     {
+        helper('alerts');
+
         if ($response instanceof RedirectResponse) {
             alerts()->session();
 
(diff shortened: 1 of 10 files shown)


michalsn:
I will rebase this PR if/when #124 will be merged.

michalsn:
Rebased.

neznaika0 (commented):
I've looked through, but I don't understand HTMX/JS yet

app/Filters/AlertsFilter.php
```diff
@@ -26,6 +26,8 @@ public function before(RequestInterface $request, $arguments = null)
      */
     public function after(RequestInterface $request, ResponseInterface $response, $arguments = null)
     {
+        helper('alerts');
```
neznaika0, line 29: I have already prepared a PR)

neznaika0: I would remove phpdoc. It contradicts the interface. This is a question for the framework. Need to either fix `void` to `null` in the interface
```php
    /**
     * @param array|null $arguments
     *
     * @return mixed
     */
```

datamweb (approved):
LGTM!
@michalsn  thank you!

lonnieezell (approved):

PR #128, opened 2023-11-06
style: String config value to FQCN

neznaika0:
FQCN more friendly and gives hints of types

### DIFF
--- a/app/Entities/UserDelete.php
+++ b/app/Entities/UserDelete.php
@@ -3,6 +3,7 @@
 namespace App\Entities;
 
 use CodeIgniter\Entity\Entity;
+use Config\Forum;
 
 class UserDelete extends Entity
 {
@@ -17,7 +18,7 @@ class UserDelete extends Entity
      */
     public function restoreLink(): string
     {
-        $days = config('Forum')->accountDeleteAfter * DAY;
+        $days = config(Forum::class)->accountDeleteAfter * DAY;
 
         return signedurl()
             ->setExpiration($days)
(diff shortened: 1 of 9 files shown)


michalsn:
While I agree when it comes to regular PHP files. I disagree about doing this for the view files.

neznaika0:
The view files are not PHP? They also have syntax highlighting and annoying underscores of undefined parameters...

michalsn:
I probably could have used different wording, however, I personally prefer not to use `use` in view files. 

But that's just me. I will adjust if necessary.

neznaika0:
We can wait for the other participants. Everyone has different time zones

datamweb:
@neznaika0 I thought a little about this PR. What I have noticed is that there has not been much desire to implement this method regarding Views.

I don't know what's the logic behind it, but views are not that important from the core team's point of view(maybe because the views can only be one HTML file). If you pay attention, we have [removed directory Views](https://github.com/lonnieezell/forum-example/blob/c7dd0b01eb48e3a24e6882c86218ed672337b089/phpstan.neon.dist#L13-L13) from the static analysis. This issue is also present in Shield and...

neznaika0:
Hints are required by the developer and not by the analyzer. It is more convenient for me when there is auto-completion. Especially when the object has many fields and methods. Your desire, PR is canceled

kenjis:
I prefer `::class` because I can autocomplete and jump to the definition.

neznaika0:
But **Shield** contains always `config('Auth')`, for example

kenjis:
It is because to allow override the Config by another module.
If you place `Config\Auth`, it will take precedence.
In most apps, you use `Config\Auth`, and `config(Config\Auth::class)` is no problem.

But if you don't use `Config\Auth`, and want to put it in another module like `Foo\Config\Auth`,
we need to use `config('Auth')`.

PR #129, opened 2023-11-06
Set default sort for discussion list

neznaika0:
By default, sorting is not defined for HTML, but the model uses **recent-posts**.
The initial type is incorrect **recent-threads**


michalsn (approved):
I'm not sure if we shouldn't go with `recent-threads` in [model](https://github.com/lonnieezell/forum-example/blob/develop/app/Models/ThreadModel.php#L102) as the default value, but it's a good fix for now. Thanks!

PR #131, opened 2023-11-06
Disable access to private categories

neznaika0:
Now it is acceptable to view discussions, create/edit for a thread (dropdown HTML) of a private/inactive category.

### DIFF
--- a/app/Controllers/Discussions/DiscussionController.php
+++ b/app/Controllers/Discussions/DiscussionController.php
@@ -62,9 +62,17 @@ class DiscussionController extends BaseController
 
     /**
      * Display a standard forum-style list of discussions.
+     *
+     * @throws PageNotFoundException
      */
     public function category(string $slug): string
     {
+        $activeCategory = model(CategoryModel::class)->active()->public()->findBySlug($slug);
+
+        if (! $activeCategory) {
+            throw PageNotFoundException::forPageNotFound();
+        }
+
         $table = [
             'perPage' => $this->request->getGet('perPage') ?? 20,
             'search'  => $this->request->getGet('search') ?? ['type' => 'recent-posts'],
@@ -94,7 +102,7 @@ class DiscussionController extends BaseController
                 'pager' => $threadModel->pager,
             ],
             'searchUrl'      => route_to('category', $slug),
-            'activeCategory' => model(CategoryModel::class)->findBySlug($slug),
+            'activeCategory' => $activeCategory,
         ];
 
         $data['table'] = [...$table, ...$data['table']];
--- a/app/Controllers/Discussions/ThreadController.php
+++ b/app/Controllers/Discussions/ThreadController.php
@@ -46,7 +46,7 @@ class ThreadController extends BaseController
 
         helper('form');
 
-        $categoryDropdown = model(CategoryModel::class)->findAllNestedDropdown();
+        $categoryDropdown = model(CategoryModel::class)->active()->public()->findAllNestedDropdown();
 
         if ($this->request->is('post')) {
             $validCategoryIds = array_reduce($categoryDropdown, static fn ($keys, $innerArray) => array_merge($keys, array_keys($innerArray)), []);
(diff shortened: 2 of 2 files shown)


michalsn:
Hiding private categories for everyone is not the best choice, this should rely on permissions.

neznaika0:
I haven't worked with Shield yet and haven't studied the forum in detail. I saw that you are already using such a query

datamweb:
> I haven't worked with Shield yet and haven't studied the forum in detail.

See [Policies]( https://github.com/lonnieezell/forum-example/blob/develop/_docs/policies.md)

neznaika0:
in demo mode (after start Seeder) I don't understand where what rights apply.
The rights for users in the table are empty, for categories **N;**
How do you work with rights? I roughly understand the policy, but without the initial rights it is unclear what to do.

michalsn:
In the first category, there are some permissions set in the `permissions` field. These permissions indicate who has access to the category and all its children.

neznaika0:
I understood a little how it works. Now another question. All requests to categories must have visibility settings.
So we need to: 
1. Build each request in the controller manually 
``` php
        if ($this->policy->can('categories.viewPrivate')) {
            $activeCategory = model(CategoryModel::class)->findBySlug($slug);
        } else {
            $activeCategory = model(CategoryModel::class)->public()->findBySlug($slug);
        }
```
2. Configure each request in the model, but with the introduction of auth/policies by services 
```php
        $allchildren = $this
            ->active()
            ->children();
           // check auth()->can()
            $allChildern->select(implode(', ', $selects))
            ->join('threads', 'threads.id = categories.last_thread_id')
            ->join('users', 'users.id = threads.author_id')
            ->orderBy('order', 'asc')
            ->findAll();
```
3. Create a separate method for each setting
```php
    public function findAllNested() // public and private
    public function findAllNestedPublic()
    public function findAllNestedPrivate()
```
4. Add (bool) options. Not recommend
Your suggestions? Each method has pros and cons.
Now there are practically no such checks anywhere. The rights check is not working correctly
This remark applies to the `active` parameter of the category

michalsn:
I believe the easiest way would be to leave the model as it is.

In the database seeder, there is an example of how this will work. We have: `admin.access` and `categories.moderate` permissions assigned to the first category. These permissions will determine access to a category. The permissions that give access can be chosen freely. 

I guess you could add a new method to the `Policies` class that will return `true` if the user has been granted permissions to any of the passed parameters (from the `permissions` field in the database)

That is, we return all the results for the categories, and then iterate through them and check if the user has the appropriate permissions. If not, we remove the category and all its children from the result.

We should make it quite flexible and simple because we will be checking it in many places.

neznaika0:
Does this mean using the second option? With the using of Auth/Policy services.
It is unclear how the synchronization of the `active/private` table fields and permissions will be? Do I need to add `showPrivate` to permissions or check `active/private` directly (1 or 0)?

michalsn:
I think the `private` field simply means that additional permissions are set for these categories. So I wouldn't worry about this field.

However, I could be wrong. Maybe @lonnieezell can shed some light on what the exact idea/function of these fields was. I'm not sure if we need the `private` field at all since we have to check the `permissions` field anyway.

Anyway... no, what I was proposing is not exactly the 2nd option. What I was saying is: Don't bother with any changes when it comes to the query. The query is fine as it is. You should grab all the categories and then check the permissions. There should be no "special" permission that indicates the category is private. You have to check the permissions which are declared in the `permissions` field.

lonnieezell:
Originally `categories` were `forums` in the demo I had started putting together before this picked up steam. The intent was to allow for private discussion groups, either for moderator/admin use or even for allowing users to create their own groups. While I'd still love to allow that type of feature in the future, it's not needed in the `category` system that we have, I don't think, especially with the permissions. So I think that column can be safely removed. 

If/when we add groups in the future I think they probably need to be their own entity instead of relying on categories. 

I agree with the solution that @michalsn was putting out in his last post.

neznaika0:
Sadly, the language barrier and the level of knowledge limits my help. I think until someone adds an example, I'll stop.

michalsn:
Resolved by #217

PR #135, opened 2023-11-07
2FA

### LINKED ISSUE #121: Inactive Members should be pruned after a configurable inactive time.
When a user has been inactive for an extended length of time (measure in years) the account and all data should be removed, after a couple of warning emails. 

- new Tasks that run weekly
  - notify users 30 days from deletion date
  - notify users 7 days from deletion date
  - notify users 3 days from deletion date
  - delete users  and their data that have been inactive for the set amount of months
- config item to specify how many months should be considered for deletion

EDIT: Maybe they should just be soft-deleted? That protects the old account but still allows it to be restored somehow? Open for discussion on this. 

reference: https://blog.google/technology/safety-security/updating-our-inactive-account-policies/

michalsn:
This PR adds 2FA for user accounts. Any user can enable it if they want. 2FA is forced when a user's last login is older than 12 months (this can be configured via `Config\Forum::force2faAfter`).

Closes #121


lonnieezell (commented):
First - this looks great. Thanks!

However, I don't see how this forces 2FA after 12 months. I see the check within the 2FA email action, but this wouldn't run for someone without 2FA would it? I think that's pretty crucial to make this feature work.

michalsn:
This is done here: https://github.com/lonnieezell/forum-example/pull/135/files#diff-4625d7a0608ad6e19b7bec5e139a0554445eea45ad849742e2269ae675c76c90R29-R41

- First, we check that 2FA is enabled and follow it
- Then, we check if the last successful login was older than X (12 months). And if this is true, we force 2FA
- If none of the above is true, we simply let the user in

lonnieezell:
Ah, I get it now. Since that's set it will always run - which is the part I was thinking clearly on. Then, if they have a `two_factor_auth_email` value set, meaning they've opted in, it always runs, otherwise does the appropriate check. 

That makes sense. Thanks.

lonnieezell (approved):

michalsn:
Yes, exactly.

I just realized that I should probably handle a scenario where a user created an account but never logged in. This case should force a 2FA after 12 months of account creation.

neznaika0:
Yes, this is a special case. After registration there is no entry in the authorization history.

michalsn:
This one is ready.

app/Controllers/Account/SecurityController.php
```diff
@@ -122,6 +122,59 @@ public function changePassword()
...
     }
 
+    /**
+     * Enables / disables 2FA (email) for the account.
+     */
+    public function twoFactorAuthEmail()
+    {
+        if (! $this->policy->can('users.twoFactorAuthEmail', auth()->user())) {
+            // user is allowed NOT to change 2FA settings
+            return $this->policy->deny('You are not allowed to change the 2FA settings for this user.');
```
datamweb, line 132: It seems it's time to use `alerts()`.

app/Views/account/security/_two_factor_auth_email.php
```diff
@@ -0,0 +1,45 @@
+<div x-data="{ open: <?= ! empty($open) && $open ? 'true' : 'false' ?> }" id="two-factor-auth-email-form">
+    <div class="flex justify-between">
+        <div>
+            <h3 class="font-bold">2FA <small>(Two-Factor Authentication)</small></h3>
```
datamweb, line 4: It was good to mention the word Email, maybe we will use Google Authenticator in the future.

app/Libraries/Authentication/Actions/TwoFactorAuthEmail.php
```diff
@@ -0,0 +1,44 @@
...
+    public function createIdentity(User $user): string
+    {
+        if ($user->two_factor_auth_email) {
+            return parent::createIdentity($user);
+        }
+
+        // Last login date or account creation date
+        $date = $user->lastLogin()?->date ?? $user->created_at;
+        if ($date->difference(Time::now())->getMonths() >= config(Forum::class)->force2faAfter
+            || model(UserIdentityModel::class)->getIdentityByType($user, $this->getType()) !== null) {
```
datamweb, line 38: `getType()` It would be better to set a new name for it. It is currently `email_2fa`, but in fact you have customized the action.

datamweb (commented):
I'm late, but it works well.

Screenshot 2023-11-09 180616

app/Controllers/Account/SecurityController.php
michalsn: Not sure about it. I think it's good as it is. People should see this info only if they manually manipulate the link.

app/Views/account/security/_two_factor_auth_email.php
michalsn: There is a description below, but if we add more options then this definitely will have to change.

app/Libraries/Authentication/Actions/TwoFactorAuthEmail.php
michalsn: Good point.

PR #137, opened 2023-11-09
Initial vite integration

lonnieezell:
This removes ParcelJS and switches our asset built pipeline over to [Vite](https://vitejs.dev/). This should hopefully give us a more stable experience. A couple of notes on it: 

1) This only works for development environments currently. While `composer` has a script to build the production assets, it has not been configured to output to the correct location. In addition, the new `Vite` class needs to be updated to locate the built manifest and retrieve the proper URL from that. 
2) HTMX is loaded and working correctly, but I'm seeing an error with cache history all of the time and I can't pin down the source of it just yet. 

While fixing this, I ran into an issue where Chrome wasn't recognizing the `mask` classes without being prefixed. This seems to be a known bug, though canIUse supposedly says it has been able to for quite some time. The `autoprefixer` plugin for PostCSS uses canIUse to determine what needs prefixing, I believe.

Finally - the site is running slow for me now. The time seems to be within the backend portion, so I don't believe it has anything to with this conversion, but is something to track down.


app/Entities/Post.php
```diff
@@ -57,6 +57,6 @@ public function setReplyTo(?string $value = null)
 
     public function isMarkedAsDeleted(): bool
     {
-        return $this->attributes['marked_as_deleted'] !== null;
+        return $this->attributes['marked_as_deleted'] ?? null !== null;
```
neznaika0: Always `null`. Why this change? If `marked_as_deleted` may not be initialized, then is better?
```php
protected $attributes = [
        'marked_as_deleted' => null
];
```
Although the date cast should apply null itself..

michalsn:
> Finally - the site is running slow for me now. The time seems to be within the backend portion, so I don't believe it has anything to with this conversion, but is something to track down.

That's strange because I have noticed a speed boost, that is noticeable to me. There is no difference in request time per se (they were always fast - worst case scenario a little over 300ms) but the content swap when we were loading a full page with HTMX (with HTML tag etc.) was a bit slow with Percel. With Vite I feel no delay in content swap.

app/Entities/Post.php
michalsn: This is probably just a workaround for an unmigrated database? At least it seems like it.

lonnieezell: That's what I figured, too. Pretty sure I ran migrations at that point again, but we can remove this check for now.

michalsn (approved):
I'm not sure if you want to add more features to this PR, but it is already usable.

lonnieezell:
> I'm not sure if you want to add more features to this PR, but it is already usable.

I don't currently. Obviously, more needs to be done before the project gets a release, but not needed for current development. I just need to take a pass and make Rector, Stan, and Co. happy tonight.

PR #146, opened 2023-11-09
2fa update

michalsn:
This PR implements a suggestion from the code review and other small things.

* ~The `$type` for the `TwoFactorAuthEmail` action has been changed~
* We're making sure we only display options that users are allowed to use on the `Security` page
* The `policy` helper function was fixed

### DIFF
--- a/app/Common.php
+++ b/app/Common.php
@@ -21,6 +21,6 @@ if (! function_exists('policy')) {
      */
     function policy(string $permission, mixed ...$arguments)
     {
-        return service('policy')->check($permission, ...$arguments);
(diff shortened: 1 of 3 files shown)


lonnieezell (approved):

datamweb (changes requested):
@michalsn, it seems that due to the extends Email2FA , the result that we expected is not obtained.

Therefore, these PR should not be merged.

michalsn:
@datamweb I forgot to check it and indeed it doesn't work... we went back to the original type though.

Thanks for double-checking.

datamweb (approved):
Thanks!

PR #147, opened 2023-11-10
Shield b6 to b7

datamweb:
This PR is trying to upgrade the Shield version from B6 to B7.

Sorry for the long delay.

### DIFF
--- a/app/Config/Auth.php
+++ b/app/Config/Auth.php
@@ -487,8 +487,8 @@ class Auth extends ShieldAuth
     {
         return match (true) {
             str_starts_with($url, 'http://') || str_starts_with($url, 'https://') => $url,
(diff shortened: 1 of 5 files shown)


app/Config/Auth.php
```diff
@@ -224,7 +271,7 @@ class Auth extends ShieldAuth
      */
     public array $validFields = [
         'email',
-        'username',
+        // 'username',
```
kenjis, line 274: Is this okay?

datamweb: Currently yes, no problem. We currently have email/password login.
Probably originally came from Bonfire2.

lonnieezell (approved):
No worries about the delay! Thanks for doing this. I appreciate it

PR #149, opened 2023-11-10
chore: add phpstan-codeigniter

kenjis:
- add phpstan-codeigniter
- refactor
- composer update

### DIFF
--- a/app/Common.php
+++ b/app/Common.php
@@ -16,10 +16,8 @@
 if (! function_exists('policy')) {
     /**
      * A convenience method for accessing the Policy service.
-     *
-     * @return mixed
      */
-    function policy(string $permission, mixed ...$arguments)
+    function policy(string $permission, mixed ...$arguments): bool
     {
         return service('policy')->can($permission, ...$arguments);
     }
--- a/app/Config/Services.php
+++ b/app/Config/Services.php
@@ -22,7 +22,7 @@ use CodeIgniter\Config\BaseService;
  */
 class Services extends BaseService
 {
-    public static function policy($getShared = true)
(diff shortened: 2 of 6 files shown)


lonnieezell:
I have to be honest. I'm a little leery of this one. I feel like that can be over strict in places and make submitting PRs more painful than it needs to be. This is a side project that is intended to be a fun side project. I don't want it to become what CI is where 1/2 of the time is spent managing the quality tools. Yes, that's probably harsh on my part but it feels that way some days. 

I'll go ahead and allow it here, but I'll reserve the same right as I told @michalsn  previously - if the tools block me from a PR because I don't have the proper generic listed in the docblock, I make no promises how long it will stick around :)

lonnieezell (approved):

michalsn (approved):
I have less and less hair on my head, so I hope this will not contribute to losing more 😅 😆 

But I choose to be optimistic 🤞

kenjis:
Complicated type specifications should become unnecessary the more well-organized the code base (including the framework/libraries) type specifications are.

That said, we all have struggled with type specifications in the past, so I understand your concern.
So, if type errors are preventing your PR, feel free to suppress the errors.

phpstan-codeigniter is intended to make type inference in CI4 more precise, not to make type specification more strict. However, it is possible that erroneous errors may be reported due to missing type specifications. So, I suppressed the found false errors.

kenjis:
What I really want is simple code with simple types.
Not `@phpstan-return ($id is int|string ? array<int|string, float|int|null|string>|object|null : list<array<int|string, float|int|null|string>|object>)`, but `@return Entity`

Issue #122, opened 2023-11-02
Help area


lonnieezell:
A section of the site should be available that displays a help section. 

- The pages should be markdown files in their own folder hierarchy. 
- The folder names are used as the category they are displayed in
- The filenames can be prefixed with a number and an underscore to help order files, if needed.
- The filename is the title of the page, which should ignore the numeric prefix if it is found.

The main page should display a search form  and a selection of links to some of the pages. Probably by listing each category and the top 5 pages?

PR #154, opened 2023-11-13
feat: Help area

michalsn:
This PR adds a help area as described in #122

* For now, only sample files are included, so we can see how it's working.
* We need some styling for the markdown-generated code since we're getting simple HTML without any classes. All the content looks the same.


lonnieezell (approved):
Looks great. Nice job with your MarkdownPages library, too!

michalsn:
Thanks!

PR #155, opened 2023-11-14
Account avatar support

lonnieezell:
- Includes new Filesystem support for saving/reading files from multiple locations, like local disk, S3, etc. This is a simple wrapper around the [Flysystem library](https://flysystem.thephpleague.com/docs/). 
- adds Avatar to the profile form, along with the ability to delete the image from the form also. 

I know that testing is pretty slack on this. I'm not actually sure how to test uploading files in CodeIgniter. A quick search and a few people are saying it's not possible, so that's something we need to look into for the core code. Providing a simple way to test file uploads. 

Once we have that we can update the tests for this.


app/Controllers/Account/AccountController.php
```diff
@@ -86,25 +88,63 @@ public function profile()
...
             $user = auth()->user();
             $user->fill($this->validator->getValidated());
 
+            // If we have a new avatar, delete the old one first,
+            // then save the new one to the configured Storage.
+            if ($this->request->getFile('avatar')->isValid()) {
+                try {
+                    $user->deleteAvatar();
+                    $user->avatar = $user->saveAvatar($this->request->getFile('avatar'));
+                } catch (FileNotFoundException $e) {
```
michalsn:
```suggestion
                } catch (FilesystemException $e) {
```

app/Controllers/Account/AccountController.php
```diff
@@ -86,25 +88,63 @@ public function profile()
...
+
+        try {
+            $user->deleteAvatar();
+
+            if (model(UserModel::class)->save($user)) {
+                alerts()->set('success', 'Your avatar has been deleted');
+            } else {
+                alerts()->set('error', 'Something went wrong');
+            }
+        } catch (FileNotFoundException $e) {
```
michalsn:
```suggestion
        } catch (FilesystemException $e) {
```

app/Controllers/Account/AccountController.php
```diff
@@ -4,9 +4,11 @@
 
 use App\Controllers\BaseController;
 use App\Entities\NotificationSetting;
+use App\Entities\User;
 use App\Models\NotificationSettingModel;
 use App\Models\PostModel;
 use App\Models\UserModel;
+use CodeIgniter\Files\Exceptions\FileNotFoundException;
```
michalsn:
```suggestion
use League\Flysystem\FilesystemException;
```

app/Views/account/_profile.php
```diff
@@ -78,6 +82,36 @@
...
+                    <?php endif ?>
+                </div>
+
+                <div>
+                    <?= form_upload([
+                        'name' => 'avatar',
+                        'class' => 'file-upload w-full',
+                        'accept' => 'image/png,image/jpeg,image/jpg',
+                    ]) ?>
+                    <p class="text-sm opacity-50 mt-2">Maximum file size: <?= max_upload_size() ?></p>
```
michalsn: In my dev env, this will display 512 MB which is misleading since we can only upload 1MB.

app/Controllers/Account/AccountController.php
```diff
@@ -86,25 +88,63 @@ public function profile()
             'location'  => ['permit_empty', 'string', 'max_length[255]'],
             'company'   => ['permit_empty', 'string', 'max_length[255]'],
             'signature' => ['permit_empty', 'string', 'max_length[255]'],
+            'avatar'    => ['permit_empty', 'uploaded[avatar]', 'mime_in[avatar,image/jpg,image/jpeg,image/png]', 'max_size[avatar,1024]'],
```
michalsn: Apparently `permit_empty` cannot be used with file validation, because it causes the file validation to be completely ignored.

michalsn (commented):
I finally had time to "click" around this PR. Looks really nice.

app/Controllers/Account/AccountController.php
lonnieezell:
```suggestion
            'avatar'    => ['uploaded[avatar]', 'mime_in[avatar,image/jpg,image/jpeg,image/png]', 'max_size[avatar,1024]'],
```

app/Views/account/_profile.php
lonnieezell: Doh! I think I was picking `max` instead of `min`. I'll double check that.

michalsn: No, that's not it. The thing is that the max file size is defined in the validation rule. And we're displaying the "min" allowed size by the php.

composer.json
```diff
@@ -12,6 +12,9 @@
     "codeigniter4/shield": "^1.0@beta",
     "codeigniter4/tasks": "dev-develop",
     "league/commonmark": "^2.3",
+    "league/flysystem": "^3.18",
```
datamweb, line 15: Hmm, looks like you need require `ext-fileinfo` .

datamweb:
Screenshot 2023-11-18 000641

Except for the mentioned + conflicts , everything looks good.

lonnieezell:
Alright - went to fix the file size issue and ended up with a few other tweaks: 

- use the ImageUpload config file, which I'd forgotten about. 
- modified to respect fileSize value in config file if greater than 0. If 0 then will examine the system to determine size. 
- Updated the view to use that same fileSize value
- Fixed a bug when an error happened it was returning the entire page inside the form area.

app/Views/account/_profile.php
lonnieezell: Gotcha! Should be fixed up in the latest push.

michalsn (approved):
Looks great!

PR #158, opened 2023-11-15
Load htmx.js

michalsn:
This PR fixes an error:
```
Uncaught ReferenceError: htmx is not defined
```
when trying to sort discussions by type (dropdown).

Also `htmx.org/dist/ext/preload.js` seemed not to be used in the code, so I removed it.

### DIFF
--- a/themes/default/js/app.js
+++ b/themes/default/js/app.js
@@ -1,4 +1,4 @@
-import "htmx.org/dist/ext/preload.js";
+import "./htmx.js";
 import "htmx.org/dist/ext/loading-states";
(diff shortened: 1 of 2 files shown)


datamweb (approved):
LGTM!

lonnieezell (approved):
I think was leftover from a bunch of attempts to get things to work. Your solultion is actually one that was mentioned in one of the [issues](https://github.com/bigskysoftware/htmx/issues/1690) on their site, though, so I'm good with it if it solved all of the issues.

PR #167, opened 2023-11-20
New CI colored theme support with dark mode toggles

lonnieezell:
- Created a couple of theme variations to more closely mimic the colors we're using on the Shield site. These are just a few colors layered on top of the existing light/dark themes. We can add more variations as we go if we find we need it. 
- Added a theme switching system 
- Did a little cleanup of a couple of for looks. 

KNOWN ISSUES:
- The theme switching stops working when a boosted page is loaded. I think we might be using boosting a little much, honestly, especially with the category list, and that might be some of the performance issues I'm seeing with Chrome. I should note that Firefox performs flawlessly for me, though I still think loading out discussions shouldn't take as long as does. I have a feeling that's because we're rendering the markdown live on each render. Thinking we should either cache that to the database on write, or start caching each generated post.

### DIFF
--- a/app/Views/discussions/posts/_create.php
+++ b/app/Views/discussions/posts/_create.php
@@ -12,9 +12,9 @@
                 Create a new <?= empty($post_id) ? 'post' : 'reply'; ?>
             </div>
             <div x-data="{ tab: 'message' }">
-                    <div class="tabs mb-1">
-                        <a class="tab tab-lifted" :class="{ 'tab-active': tab === 'message' }" @click.prevent="tab = 'message'">Message</a>
-                        <button class="tab tab-lifted"
+                    <div role="tablist" class="tabs tabs-bordered mb-2">
+                        <a role="tab" class="tab" :class="{ 'tab-active': tab === 'message' }" @click.prevent="tab = 'message'">Message</a>
+                        <button role="tab" class="tab"
                                 :class="{ 'tab-active': tab === 'preview' }"
                                 x-on:preview-show="tab = 'preview'"
                                 hx-confirm="unset"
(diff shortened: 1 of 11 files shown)


michalsn:
Looks good!

One thing I am wondering about is... shouldn't we use: https://daisyui.com/components/theme-controller/ for theme switching? Also, it would be good if the switcher would be available for everyone not just logged-in users.

To make switcher work after boosted request we can wrap it into `htmx.on("htmx:load"...` - just like `initEditor` or `initTags` so that all binding can be recreated.

When it comes to boosting and Chrome - seems like this might be an issue: https://github.com/bigskysoftware/htmx/issues/879. I use Firefox as a default browser, but after checking, there is indeed a slower performance in Chrome. We can make a normal request for now for categories and will see.

I haven't noticed any speed issues when parsing markdown yet. What times do you get exactly? Although, at some point, we probably should save parsed markdown.

datamweb:
> One thing I am wondering about is... shouldn't we use: https://daisyui.com/components/theme-controller/ for theme switching? Also, it would be good if the switcher would be available for everyone not just logged-in users.

I agree. These two things must be considered.


In addition to the mentioned issues, I see some small problems.

Screenshot 2023-11-20 201853

Screenshot 2023-11-20 201929

Screenshot 2023-11-20 202059

michalsn (approved):
Design flops are not directly related to the code changes. I guesss they're caused by Daisy UI version bump.

The theme switcher changes are up to you. It can always be done later. The theme changes are worth to be merged.

lonnieezell:
Sorry - had other stuff going on the last few days. 

I suppose I agree on moving the toggler to where it can be accessed without being logged in. I'll handle that later and take a look at the DaisyUI component for that, which I missed before. 

My intent right now was to start going through what we have so far and tweaking the UI where needed, so that will be my focus for the next couple of weeks, I think. I can address these other issues as part of those. 

Merging what we have for now.

Issue #143, opened 2023-11-09
Users can Flag a post/thread


lonnieezell:
In a new bar at the bottom of posts and threads, there should be a button to Flag the post. When pressed, it verifies they want to flag it and presents a textarea that allows them to enter WHY they think it should be flagged. 

Once submitted, the post/thread is added to the moderation queue with the user_id of the flagger and their message. Both the submitter and the message should be available in the moderation queue.

Issue #142, opened 2023-11-09
Mods: Moderator Log


lonnieezell:
This page displays the logs of all of the actions that a moderator has taken.

Should be able to filter the logs by: 

- category
- moderator
- date range (with a few common ranges or the ability to select a specific date
- can sort columns

Issue #141, opened 2023-11-09
Mods: Moderate Posts


lonnieezell:
Within the moderator area, will display the posts that were marked for moderation so that a mod can:
     - approve the message, which removes it from moderator queue
     - deny the message, deletes the message
     - ignore the message and keep it in the queue

Should support bulk actions for the current page of results.

When a moderator/admin either approves or denies, a log entry is made reflecting this.

PR #175, opened 2023-11-22
feat: moderation

michalsn:
This PR adds moderation options as per: #140, #141, #142, and #143. I know it was broken down into smaller tasks for a reason, but I found it difficult to send code that wasn't fully functioning as a whole.

For now, any user can moderate content, this can be changed simply by removing permissions.

The general question is what is missing? Should we store the IP addresses for every user report? Should we display more info in the tables when we moderate the data?

The design is not ideal, but I had no other idea for it, and tbh I'm not good with making UI without already styled components. But that may be just an excuse for not being fully invested in the look side of this project 😅 

There are still a few things missing:

- [ ] proper removing (soft-deleting) posts/threads - which will include recalculating all the stats
- [x] tests


datamweb:
This PR is big enough, honestly I always stay away from this type of PR. But well, I took a practical test. It looks good.

> The general question is what is missing?

1. If a Moderate mistakenly approves or rejects an item, how can it be canceled?
2. I would like a notification to be sent to the reporting user, to thank and inform them of approval or rejection.
3. I saw that the user who reports is not allowed to register a duplicate report, this is great, but instead of showing the message "This resource has already been reported by you", change the button **Report**  to **Reported**?

> Should we store the IP addresses for every user report?

In my opinion, there is no need for it, users' IP is intercepted and stored by the shield.

Overall, great. Thank
Screenshot 2023-11-23 001138

lonnieezell:
Collecting my thoughts here as I'm using it: 

- After reporting a post, I feel like there should be some form of notification that you reported it. Though that might get to be  a bigger performance hit than we want to deal with it. Could be nice to say something like "You report 3 days ago". Would need to move the location of the button, but we can deal with that later. 
- The `Moderate` link does not show up in the mobile navigation. 
- It would be nice to list the number of items in the queue in the moderation sidebar. 
- Navigating to `/moderation` results in a 404. Should either be a summary page or take us to the `threads` queue, I would think 
- I feel like the `Moderation options` dropdowns should be at the bottom of the list. You're likely going to be moving top to bottom in the queue and shouldn't need to go back to the top of the list to perform an action on them. 

First thoughts - nice job! You're making me feel like a slacker lol. Everything works great. I'll tackle any UI issues I see in the near future.

app/Concerns/HasThreadsAndPosts.php
```diff
@@ -0,0 +1,72 @@
...
+namespace App\Concerns;
+
+use App\Models\PostModel;
+use App\Models\ThreadModel;
+use CodeIgniter\Entity\Entity;
+
+trait HasThreadsAndPosts
+{
+    /**
+     * Adds the author and editor to each post.
```
lonnieezell:
```suggestion
     * Adds the threads and posts to each record.
```

lonnieezell (changes requested):
The code looks good, though my eyes crossed a little after reading so many files :) 

Not sure your thoughts on my previous comment, but I think at least getting the menu item added to the mobile nav should be included in this PR.

michalsn:
All mentioned things are valid. I'll make the changes and try to add some tests.

When it comes to properly deleting posts/threads, I may work on this in a separate PR.

michalsn:
@datamweb 

> If a Moderate mistakenly approves or rejects an item, how can it be canceled?

Nope, it's a one-way ticket. It could be handled with a soft delete, but I don't feel this is worth it for now. I added a confirmation. We will see if it turns out to be annoying or not.

> I would like a notification to be sent to the reporting user, to thank and inform them of approval or rejection.

This may be added as something that can be configured. But I'm not a fan of sending email notifications about this. Maybe if we someday add PM functionality? That would work fine then.

> I saw that the user who reports is not allowed to register a duplicate report, this is great, but instead of showing the message "This resource has already been reported by you", change the button Report to Reported?

Good idea, but I want to cut the number of queries to the minimum. The number of times something will be reported will be probably very low, so... not sure if want a DB call just for "one" button.

michalsn:
@lonnieezell all suggestions should be covered now.

lonnieezell:
Looks good. Thanks!

lonnieezell (approved):

PR #180, opened 2023-11-25
Add proper delete handling for threads and posts

michalsn:
This PR adds a proper delete handling for individual Threads and Posts - with statistics calculation.

### DIFF
--- a/app/Concerns/ImpactsCategoryCounts.php
+++ b/app/Concerns/ImpactsCategoryCounts.php
@@ -3,8 +3,10 @@
 namespace App\Concerns;
 
 use App\Models\CategoryModel;
+use App\Models\PostModel;
 use App\Models\ThreadModel;
 use App\Models\UserModel;
+use Exception;
 
 trait ImpactsCategoryCounts
 {
@@ -29,6 +31,8 @@ trait ImpactsCategoryCounts
 
     /**
      * Updates the category's discussion count.
+     *
+     * @throws Exception
      */
     protected function decrementThreadCount(array $data)
     {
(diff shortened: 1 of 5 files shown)


lonnieezell (approved):

app/Concerns/ImpactsCategoryCounts.php
```diff
@@ -29,19 +31,60 @@ protected function incrementThreadCount(array $data): bool|array
...
+        // The $data['id'] is always an array on delete
+        $threads = $this->allowCallbacks(false)->onlyDeleted()->find($data['id']);
+
+        foreach ($threads as $thread) {
+            $postModel = model(PostModel::class);
+            // Soft-delete all the posts for this thread
+            $postModel->builder()
+                ->update(
+                    ['deleted_at' => $thread->deleted_at],
+                    ['thread_id' => $thread->id, 'deleted_at' => null]
```
datamweb, line 52: Shouldn't `deleted_at`  be added to `$allowedFields`?
https://github.com/lonnieezell/forum-example/blob/b4010a8c72a33e64cb9a93c9f656cb26fa387beb/app/Models/PostModel.php#L25-L27

michalsn: For now, I don't think so. There is no scenario where we will be updating this field with a model.
All stats-related stuff is made with the use of a builder instance.

datamweb (approved):

app/Concerns/ImpactsCategoryCounts.php
datamweb: Okay,

datamweb:
@michalsn thank you!

PR #183, opened 2023-11-27
Shield b7 to b8

datamweb:
resolve #177

### DIFF
--- a/app/Config/Auth.php
+++ b/app/Config/Auth.php
@@ -2,6 +2,15 @@
 
 declare(strict_types=1);
 
+/**
+ * This file is part of CodeIgniter Shield.
+ *
+ * (c) CodeIgniter Foundation <admin@codeigniter.com>
+ *
+ * For the full copyright and license information, please view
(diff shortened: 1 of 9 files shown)


datamweb:
This repo target version for refactoring set php 8.1:
https://github.com/lonnieezell/forum-example/blob/73fb69ed39513734ae3d8aed2e5209cec4ab1973/rector.php#L68
but Shield is 7.4:
https://github.com/codeigniter4/shield/blob/434e5a0e550d24199dc4f04033a872e8d3d847d7/rector.php#L83

It seems that I should skip file **app/Config/Auth.php** in Rector?

michalsn:
> It seems that I should skip file app/Config/Auth.php in Rector?

Why? There are no issues with this config file.

michalsn (approved):

kenjis (approved):

lonnieezell (approved):
Awesome. Thanks for taking the time to do this!

PR #184, opened 2023-11-27
feat: Cleanup up the main app navigation

lonnieezell:
- Switched to use the new theme controller in DaisyUI
- Updated javascript to store in a cookie so that we would not see a theme flash when boosted pages were loaded. 
- Moved the account and theme switcher to always show in toolbar even in mobile
- Added styles for menu hovers
- Removed the search field for now. We can add it back in when we need it. 
- Simplified the menu entries for both mobile and desktop views. 
- Updated the theme settings in tailwind.config.js based on new info about overriding values in existing themes.

### DIFF
--- a/app/Controllers/BaseController.php
+++ b/app/Controllers/BaseController.php
@@ -68,7 +68,7 @@ abstract class BaseController extends Controller
      */
     public function initController(RequestInterface $request, ResponseInterface $response, LoggerInterface $logger)
     {
-        $this->helpers = [...$this->helpers, 'setting', 'alerts'];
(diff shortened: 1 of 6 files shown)


michalsn (approved):
Looks great!

A side note. Not sure if the current primary "light" color is not a little too bright? The much more pleasant color to look at seems to be the color we use in the official user guide. Although it's something more falling in red color than orange. But that may just be my perception.

lonnieezell:
Agree! Changed the colors to the darker orange from the main site. Good call, thanks. 

Also updated the text and active menu colors. Somehow this all made the hover state on the dropdown menu start working again, which I was fighting with previously.

PR #192, opened 2023-11-29
feat: Make primary user-facing pages themeable.

lonnieezell:
This moves the front-end view fragments to the default theme folder. This allows new themes to completely change the look of the site by copying the default folder, renaming it, and editing the view fragments. 

This adds a new View class that extends the main View class, adding a convenience method, `$this->view()` which allows the same syntax as the base `view()` method. 

In the future it would be nice to allow it to fallback to the default theme if the file doesn't exist, but that's not really important currently.


michalsn:
Hmm... not sure if we shouldn't move everything, like the rest of the `discussions` folder too. That would allow a complete change of the look if needed... but we can always do it later.

Anyway, I like the idea.

lonnieezell:
> Hmm... not sure if we shouldn't move everything, like the rest of the discussions folder too. That would allow a complete change of the look if needed... but we can always do it later.

I was borderline on that too, honestly. At some point down the road I have a project in the back of my mind that I could build off of this. My initial thinking was that the skeleton of the page doesn't necessarily need to be completely revamped, just all of the styles and maybe small details, so it seemed like leaving the page itself in the normal directory worked. But I didn't want to have to make this move at the end of the project. 

Like you said, we can always do that change later if it seems helpful, though. 

EDIT: I was adding other pages and realized the Members page was only a single page so we would need to add that, tried it and realized it worked without any issues that I was afraid of. So I moved the discussions and account pages over also.

michalsn (approved):

PR #196, opened 2023-11-30
feat: mark post as an answer

### LINKED ISSUE #191: Discussions should be able to be answered
The author of a thread should be able to mark an reply as The Answer. Once answered: 

- it should be marked in some way to signify that it was the answer
- it should no longer show up in the discussions, when "Only Unanswered" is selected. 
- should maybe show up and be clearly marked as the answer directly below the thread and before the natural posts. It should still show up within context of the rest of the threads.

michalsn:
This PR adds the ability to mark the post as an answer by the author of the thread.

Closes #191


lonnieezell (approved):
Looks great! And you remembered a couple of items I probably would have forgotten about :)

Issue #132, opened 2023-11-06
dev: Fix Psalm recommendation


neznaika0:
I understand that development is underway, but the analyzer already has ~900 comments. If the time has not come for correction, you can leave the issue until better times.

```console
$ composer analyze
...
> psalm
Target PHP version: 8.1 (inferred from composer.json) Enabled extensions: simplexml.
Scanning files...
Analyzing files...

░░

------------------------------

       No errors found!       

------------------------------
847 other issues found.
You can display them with --show-info=true
```

lonnieezell:
Anyone can feel free to submit a PR for these if they'd like.

PR #215, opened 2023-12-07
Thread page restyle and overall theme tweaks

lonnieezell:
- Restyled the view a thread page
- Moved links on threads and posts into an "action bar" with icons for actions like add, edit, delete, report, etc. 
- Restyled answered questions
- Tweaked dark and light styles. 
- Fixed ordering of columns by a person's "handed-ness"

Known issue:

The "delete" methods for threads and posts were just stubbed out. It was going to be a bigger task to make those work correctly and update the UI correctly than should be included in this PR.

Note: I tried simplifying the "handedness" of classes into a common function, but that didn't work with Tailwind since it didn't know the style should be included in the built stylesheet because it couldn't find the styles on the page.


app/Config/AuthGroups.php
```diff
@@ -149,7 +150,7 @@ class AuthGroups extends ShieldAuthGroups
             'threads.create',
             'threads.report',
```
michalsn:
```suggestion
```
If we're replacing the reporting option with only one permission, then we don't need it anymore.

app/Controllers/Discussions/PostController.php
```diff
@@ -104,6 +104,7 @@ public function create(int $threadId, ?int $postId = null)
      */
     public function edit(int $postId)
     {
+        helper('form');
```
michalsn: I see myself adding it everywhere... maybe it's a good idea to just autoload it?

michalsn (commented):
Some things I have noticed:
* On the thread list, I can't see tags if they are assigned. After entering the thread tags are visible.
* I see no option to edit a thread or post anymore
* I get an error whenever I try to report a post - my guess is the permission name should be adjusted in the controller
* I see a reporting option for my own posts/threads as well - I refuse to self sabotage 😜

lonnieezell:
Thanks for the comments. 

What role is your user? Edit and delete should only show up for the author or an admin. I know my default user was not an admin when I started messing with that.

michalsn:
I belong to the default "user" group. I cannot edit my own posts and threads. I can't see the Edit or Delete buttons. Using the `policy` service for checking permissions should fix that.

After switching my role to "admin", I started to see everything as it should be. However, there is `htmx:targetError` when I try to edit a thread. Editing posts is working fine.

lonnieezell:
I did do most of my work while logged in as an admin. 

Refactoring the action bar into a Cell so it can be more easily tested. Better in the long run, especially when we start working on the Trust system.

app/Controllers/Discussions/PostController.php
lonnieezell: Yeah, we use this everywhere. I don't see it as being a large cost. I'll add it to the BaseController.

lonnieezell:
Refactor of the Action Bar is complete.

lonnieezell:
First - I ran `cs-fix` and `rector-fix` before pushing this code up. Bastards. 

Second, and more important: @michalsn how would you feel if I ripped Psalm out? I've been thinking for a bit that I don't know that it's worth the squeeze when you compare what it catches vs how much BS you have to do and write to make the analyzer happy that really doesn't impact us developers. 

Though, to be fair, it's PHPStan I don't really like today :)

michalsn:
Oh... I feel the frustration over Psalm - I have been there and struggled too. You have to rename the cell view to: `action_bar_cell.php` (and declare a custom view name in the cell class) so it catches the rule: https://github.com/lonnieezell/forum-example/blob/develop/phpunit.xml.dist#L29 - it will solve the issue.

For me, most of the time Psalm is painless, so unless you are really determined to ditch it, I would keep it around.

michalsn (commented):
Probably the last part of the feedback.

* As an admin I can report posts and threads, but as a user I cannot. I'm guessing this is the reason: https://github.com/lonnieezell/forum-example/blob/develop/app/Controllers/Discussions/ReportController.php#L25 because we have now replaced it with `content.report`... however, I do not know why this check passes when I am an admin.
* The last thing is the tags in the thread list. I can't see them even if they are assigned to a thread. I see them only after entering the thread. Is this the way it is supposed to be?

michalsn:
ad. 1 - I see it now. It's all just because we check the "old" permission. Easy fix.

lonnieezell:
>  You have to rename the cell view to: action_bar_cell.php (and declare a custom view name in the cell class) so it catches the rule... 

Why does CI have it setup like that? I'm guessing that's old and never got switched over after we switched up how we rename views in Cells? If you use our command to generate the rule it should not get caught by our robotic overlords. But I'll do that for now. 

> For me, most of the time Psalm is painless, so unless you are really determined to ditch it, I would keep it around.

My problem wasn't so much this, but I based on #132 I went through to start cleaning it up a week or two ago and there were like 1800 errors, or something ridiculous. I spent a  little while starting to wade through, but just realized I have zero interest in teaching it how to read my code. :) I know it has some benefits, but I get so exhausted watching the issues come through from CI where it feels like 1/3 of them is us changing stuff just to make our tools happy, which is a huge loss in doing something worthwhile and the gain doesn't feel like much, especially with all of the other tooling we have in place. Maybe I'm just old and grumpy :) 

> The last thing is the tags in the thread list. I can't see them even if they are assigned to a thread. I see them only after entering the thread. Is this the way it is supposed to be?

I did remove them from the list just to clean things up. I am not convinced they are really helpful on that screen as it is. However, I've also been thinking that it would be nice if that list was more helpful and you could do more from that screen instead of having to click through to other pages. Still debating it but I might end up making it so they can be expanded so you can read the article in place, and maybe the latest few posts, or something. Not sure, but that page feels lacking to me.  I guess it feels too static and that we could take a few cues from social media to improve it instead of just being a forum.

michalsn:
> Why does CI have it setup like that? I'm guessing that's old and never got switched over after we switched up how we rename views in Cells? If you use our command to generate the rule it should not get caught by our robotic overlords. But I'll do that for now.

Yes, exactly. I keep renaming view names for cells since I don't know how else I can automatically set a rule to exclude them from the PHPUnit scan.

> My problem wasn't so much this, but I based on https://github.com/lonnieezell/forum-example/issues/132 I went through to start cleaning it up a week or two ago and there were like 1800 errors, or something ridiculous.

Fair enough, I wasn't paying much attention to it since I thought it was related to the `errorLevel` we set in the `psalm.xml`. And depending on the level we set, it will be treated as an error or simply an issue. If that is not the case, then it sucks 😜 Anyway, I'm fine with removing Psalm.

> I did remove them from the list just to clean things up. I am not convinced they are really helpful on that screen as it is. 

Okay, I get it.

> However, I've also been thinking that it would be nice if that list was more helpful and you could do more from that screen instead of having to click through to other pages. 

Yes, we can definitely experiment with this.

michalsn (approved):
Everything is starting to look a lot nicer!

michalsn:
> Still debating it but I might end up making it so they can be expanded so you can read the article in place, and maybe the latest few posts, or something. Not sure, but that page feels lacking to me. I guess it feels too static and that we could take a few cues from social media to improve it instead of just being a forum.

We could add an option to bookmark the thread, and then the latest posts would show up after login, however, this is a feature only for logged-in users - not everyone.

lonnieezell:
> Fair enough, I wasn't paying much attention to it since I thought it was related to the errorLevel we set in the psalm.xml. And depending on the level we set, it will be treated as an error or simply an issue. If that is not the case, then it sucks 😜 Anyway, I'm fine with removing Psalm.

That's a fair point and I'm not sure if all of the other Info entries were what would be included at other levels. That might be the case. I'll probably remove it anyway. I don't see the need for 2 static analyses tools. 

I'm on a bit of a "how can we simplify everything" kick right now. :)

lonnieezell:
> We could add an option to bookmark the thread, and then the latest posts would show up after login, however, this is a feature only for logged-in users - not everyone.

Definitely something to consider. 

I've also had an idea for a couple of years as I've played with forum software that it would be nice if the page you hit after logging in was more of a social network style feed, where it would know what you're watching/interacting with/etc and have the default view sorted for you. That sounds more like something that could be done down the road and easily added as another view on the main discussions list page.

PR #217, opened 2023-12-09
Check permissions for categories

michalsn:
This PR adds permission checking for categories.

* Database field `categories.private` was removed since it's not used
* The `Policy` class has an additional method: `checkCategoryPermissions`, to check if the current user has access to the given category. At first, I wanted to add permission, like `posts.view` etc, but since we require in `can()` method for the user to be logged in, I had to add another way to check this.
* We carefully check permissions. If a moderator has the right to edit a thread but does not have access to threads in some category, then he cannot edit (and see) such threads
* I added an intermediate layer between the controller and model for Categories - `CategoryManager`. So that the controller does not have too much code, and the model only performs queries to the database and is not responsible for formatting the results. This gives us a bit better separation of concerns. Please let me know if you like the idea. Since the amount of code is increasing this may be a remedy for the fat model/fat controller approach.

This PR will be rebased after #215 is merged.


app/Common.php
```diff
@@ -36,3 +39,13 @@ function vite(string|array $path): string
 
     return $vite->links($path);
 }
+
+if (! function_exists('manager')) {
```
lonnieezell, line 43: I think it probably would be okay to do the manager simply and just use `new CategoryManager()`. I don't see it as being something that we'd need to worry about a singleton for, or need to override for this project, so a service wouldn't really be needed either. 

We can keep it this way, it just seems overkill, personally.

app/Database/Seeds/SampleDataSeeder.php
```diff
@@ -13,42 +13,42 @@
...
-        'Learn More'                         => ['parent' => 'External Resources', 'order' => 2, 'active' => 1, 'private' => 0],
-        'Jobs'                               => ['parent' => 'External Resources', 'order' => 3, 'active' => 1, 'private' => 0],
-        'Addins'                             => ['parent' => 'External Resources', 'order' => 4, 'active' => 1, 'private' => 0],
-        'Archived Discussions'               => ['parent' => null, 'order' => 7, 'active' => 1, 'private' => 0],
-        'Archived General Discussion'        => ['parent' => 'Archived Discussions', 'order' => 1, 'active' => 1, 'private' => 0],
-        'Archived Libraries & Helpers'       => ['parent' => 'Archived Discussions', 'order' => 2, 'active' => 1, 'private' => 0],
-        'Archived Development & Programming' => ['parent' => 'Archived Discussions', 'order' => 3, 'active' => 1, 'private' => 0],
+        'Operations'                         => ['parent' => null, 'order' => 1, 'active' => 1, 'permissions' => ['admin.access', 'categories.moderate']],
+        'Framework Planning'                 => ['parent' => 'Operations', 'order' => 1, 'active' => 1, 'permissions' => null],
```
lonnieezell, line 17: If we're mimicing the current forums, then the next 3 categories would also be restricted.

app/Common.php
lonnieezell: I might be wrong here, actually. I see that having it in the Policy makes it where we could need a singleton.

lonnieezell (approved):

app/Common.php
michalsn: Yes, I thought so too.

app/Database/Seeds/SampleDataSeeder.php
michalsn: If a child category does not have permissions, it will inherit them from the parent category. I thought this would make sense to work that way.

michalsn:
Rebased.

app/Database/Seeds/SampleDataSeeder.php
lonnieezell: Oh, that makes sense. Thanks.

PR #232, opened 2023-12-22
Style the auth screens

lonnieezell:
- Updated Shield to `dev-develop` to take advantage of new namespace feature for auth routes
- Published views and restyled them
- Created new skeleton auth controllers in order to take advantage of our theme system.


michalsn:
I get an error when loading the `login` or `register` page.

> Invalid file: "auth.php"

lonnieezell:
Sounds like you need to do `composer update`. I changed Shield to `dev-develop` to take advantage of a new `namespaces` feature I just put in for `auth()->routes()`.

michalsn:
Yes, you're right. This looks great!

michalsn:
This PHPStan error is strange... I guess the `$users[$setting->user_id]` becomes a problem? Maybe I will check it tomorrow, but in the worst-case scenario, we can just suppress it.

michalsn (approved):

lonnieezell:
Yeah, I wasn't sure about that one either. Couldn't figure out a quick fix. But cool. I was hoping you'd say we could fix it later :)

PR #254, opened 2024-01-09
Admin User Settings page

lonnieezell:
Adds a user settings page to the admin to set basic Shield settings.


lonnieezell:
Hm. Not sure what's up with that Rector error, unless me upgrading all of the packages gave us  an incompatible version or something?

michalsn (approved):
Nice, the only issue is that I can't go back to the admin dashboard from the "Users Settings" page.

I have no idea what's up with Rector, we get similar errors on other repos too 🤷‍♂️ I guess other packages affect Rector since we have a fixed version set in the `composer.json`.

lonnieezell:
> Nice, the only issue is that I can't go back to the admin dashboard from the "Users Settings" page.

Ah, good call. I'll add breadcrumbs or something in future updates. Thanks.

PR #256, opened 2024-01-09
fix: 2fa

### LINKED ISSUE #255: Force2FA setting isn't forcing 2FA after login
The value appears to be saved correctly with settings but doesn't seem to be used by Shield. 

I only realized this just after merging the user settings branch in. 

Should force a 2fa check after login when checked.

michalsn:
This PR fixes issues with 2FA.

We use `Email2FA` class when the 2FA is enabled, and `TwoFactorAuthEmail` when 2FA is disabled.

Fixes #255

### DIFF
--- a/app/Concerns/ThemeRenderer.php
+++ b/app/Concerns/ThemeRenderer.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Concerns;
+
+trait ThemeRenderer
+{
+    /**
+     * Render a view file.
+     *
+     * Must be used in order to utilize the theme system.
+     */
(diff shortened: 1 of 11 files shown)


lonnieezell (approved):
Looks great, thanks. And I think this is a good solution - allows admins to enforce across the board if needed, otherwise gives the users a choice. 

The only thing I can think of that might benefit here is disabling and adding a note to the 2FA setting on the Account/Notifcations page.

michalsn:
Something like this would work?

image

lonnieezell:
> Something like this would work?

Yup! Works great.

PR #280, opened 2024-02-13
Trust levels - the beginning

lonnieezell:
- Config file to define trust levels/requirements/actions/etc
- `canTrustTo()` method on the User entity for checking trust actions.
- Trust Levels settings page in the admin. 

Note: the actions/requirements for these levels are not complete, but are included as a starting point. This will change as actual trust level restrictions are implemented in the code.

### DIFF
--- a/app/Controllers/Admin/Settings/TrustLevelsController.php
+++ b/app/Controllers/Admin/Settings/TrustLevelsController.php
@@ -3,9 +3,7 @@
 namespace App\Controllers\Admin\Settings;
 
 use App\Controllers\AdminController;
-use CodeIgniter\HTTP\ResponseInterface;
 use InvalidArgumentException;
-use RuntimeException;
 
 class TrustLevelsController extends AdminController
 {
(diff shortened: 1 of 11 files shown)


app/Controllers/Admin/Settings/TrustLevelsController.php
```diff
@@ -0,0 +1,74 @@
...
+use InvalidArgumentException;
+
+class TrustLevelsController extends AdminController
+{
+    /**
+     * Display the trust levels settings page.
+     */
+    public function index()
+    {
+        if ($this->request->getMethod() === 'post') {
```
datamweb:
```suggestion
        if ($this->request->is('post')) {
```
Available since version 4.3. Of course, I know that you are not interested in `is()`.

tests/Controllers/Admin/TrustLevelsSettingsTest.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Controllers\Admin;
```
datamweb:
```suggestion
namespace Tests\Controllers\Admin;
```

app/Controllers/Admin/Settings/TrustLevelsController.php
lonnieezell: Never said I wasn't interested. I knew getMethod had a deprecation but couldn't remember what the alternate was. And completely forgot about `is()` so thanks for reminding me!

tests/Controllers/Admin/TrustLevelsSettingsTest.php
lonnieezell: Good catch. Thanks!

PR #303, opened 2024-03-12
Signature trust

lonnieezell:
- Display author signatures on posts and threads
- Render the signatures with Markdown
- Ensure all links for post/thread content and signatures are nofollow and open in new tab
- Trust levels define whether links can be displayed within a user's signature
- setup caching for post/thread bodies and signatures

### DIFF
--- a/app/Concerns/RendersContent.php
+++ b/app/Concerns/RendersContent.php
@@ -3,6 +3,7 @@
 namespace App\Concerns;
 
 use App\Libraries\TextFormatter;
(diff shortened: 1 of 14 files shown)


lonnieezell (commented):
Thanks for helping out! 

Since this is all handled in the CodeIgniter user guide, I think it would make more sense to just link this step to this page: https://codeigniter.com/user_guide/installation/running.html

That way we don't have to repeat the same information.

PR #339, opened 2024-04-18
fix: load text_helper

dimtrovich:
the `random_string` function is in the helper text, but this is not loaded at any time, which can generate an error when executing the seed

image


datamweb (approved):
LGTM! thanks.

lonnieezell (approved):
Good catch, thanks!

PR #340, opened 2024-04-18
chore: fix github action run with custom config for code style

datamweb:
```console
In Finder.php line 670:

  [LogicException]                                                             
  You must call one of in() or append() methods before iterating over a Finde  
  r.                                                                           
```

See https://github.com/lonnieezell/forum-example/actions/runs/8735608318/job/23971875926


lonnieezell (approved):
Thanks!

PR #390, opened 2024-09-01
fix: Restore Routes file

neznaika0:
Initially, someone deleted the default **app/Config/Routes.php**. There is a constant error in the log. 

`WARNING - 2024-09-01 19:56:39 --> Routes file not found: "/home/aleksandr/www/codeigniter/forum-example/app/Config/Routes.php"`

This file should be (even empty), the system reads it.

### DIFF
--- a/app/Config/Routes.php
+++ b/app/Config/Routes.php
@@ -1,0 +1,3 @@
+<?php
+
+declare(strict_types=1);


lonnieezell:
Right. Routes were moved out to `/app` directory. And `app/Config/Routing.php` [was updated](https://github.com/lonnieezell/forum-example/blob/develop/app/Config/Routing.php#L28). 

I guess there's still someplace in CI that is looking for that file, though. Bummer. I'm ok with this as a workaround. Thanks.

lonnieezell (approved):

neznaika0:
I've seen the changes. The [Routing](https://github.com/codeigniter4/CodeIgniter4/blob/c326cd1d98de2c50e90156347a2247c6e4cb7b2c/system/Config/Routing.php#L32) file hopes on default routes. 
Config arrays are merged.

neznaika0:
About the test error.
It seems there is an error when creating the test. If you run the `composer test
./tests/Policies/ThreadPolicyTest.php` - the result is correct

**EDIT:** Yes. The test is written incorrectly. If you restart it multiple times, the error disappears periodically. Depends on the row from the DB

PR #392, opened 2024-09-04
Fix after clean install and misc

neznaika0:
Sorry, for one PR. It's easier to apply the changes this way. The excess can be returned to its original state.

these are general fixes that are found after a clean installation and exemplary user behavior.

- [x] fix: Add `id` for HTMX action
"Edit Thread" button does not work
- [x] fix: Add LEFT JOIN instead of JOIN
If there are no Categories or Threads/Posts inside, the list of Categories is empty. It should be displayed regardless of Trends/Posts
- [x] fix: Add URL for home page
Clicking on the Logo/Sitename does not lead anywhere. Annoy
- [x] fix: Сase when no categories
If there are no categories in the database, exceptions will occur. You need to add empty arrays by default
- [x] fix: Configure cache handler
The initial setup does not provide for the use of a Cache. Corrected .env
- [x] fix: Delete unnecessary namespaces
- [x] fix: Bad encoding DOMDocument 
The texts were displayed incorrectly after formatting. See https://www.php.net/manual/en/domdocument.loadhtml.php#74777
EDIT: HTML-ENTITIES will be deprecated. See https://www.php.net/manual/en/function.mb-convert-encoding.php#127529
- [x] Hide pagination if total pages = 1


app/Models/CategoryModel.php
```diff
@@ -87,17 +91,17 @@ public function findAllNested(array $categoryIds): array
             ->parents()
             ->whereIn("{$this->table}.id", $categoryIds)
             ->select(implode(', ', $selects))
-            ->join('threads', 'threads.id = categories.last_thread_id')
-            ->join('users', 'users.id = threads.author_id')
+            ->join('threads', 'threads.id = categories.last_thread_id', 'LEFT')
```
lonnieezell, line 94: Where are these used? If it's only for the category list in the sidebar then having only categories with posts in them is desirable so that users don't waste time clicking into categories that don't have any content. 

If it's used elsewhere, then I agree with this.

neznaika0: This is very important :smile: 
You will not be able to create a new topic in an empty Category. Select menu is empty.
Everything works for you because you are using a Seeder or an already filled database

lonnieezell: Ok, that's what I was wanting to verify. :) The rest looks good. Approved.

lonnieezell (approved):

neznaika0:
Why is the cache permanent?

https://github.com/lonnieezell/forum-example/blob/afa76b0138e32d5f36acc9199c96a9856c3bb1b9/app/Managers/CategoryManager.php#L31

lonnieezell:
> Why is the cache permanent?
> 
> https://github.com/lonnieezell/forum-example/blob/afa76b0138e32d5f36acc9199c96a9856c3bb1b9/app/Managers/CategoryManager.php#L31

Without digging back through the code, I would assume it's because they don't change often. You would destroy the cache only when permissions change.

neznaika0:
Okay. I've already forgotten what I was trying to fix, too. I think it's worth testing and checking the PR again.

neznaika0:
The error appears sometimes. I don't think you should look at her right now.
I ran on PHP **8.4** and there are a lot of errors. And CI **v4.6** already requires a lot of fixes.

Interest is lost. I hope someone will continue to develop :cry:

neznaika0:
@lonnieezell I updated and checked on version 4.6.3, no problems were found, the tests were successful. Can you merge them?

lonnieezell:
Thanks!

PR #398, opened 2025-11-30
fix: Dark mode switcher

### LINKED ISSUE #395: bug: Theme switcher not working
I ran a test sample, with no data (just tables). I installed composer and npm packages from JSON.
The dark theme switch does not work on the main page. It can switch sometimes. 

I believe this is a problem with overwriting Cookies - always `light`.

neznaika0:
Fixes #395 
I don't know much about JS, but I managed to set up the switch. Can someone explain the duplication of scripts for themes?


lonnieezell (approved):
I pulled the code down and was not able to replicate the previous issue as I had seen it. Looks good to me.

neznaika0:
Thanks, I tried not to break it. While this is a demo version, I think it can be improved later.

PR #399, opened 2025-11-30
fix: Improved work with Post

neznaika0:
- [x] Missing attributes have been added for saving via the Model. It was possible to insert only through QueryBuilder.
- [x] Added a test to check for changes to deleted Post.
- [x] 100% coverage `ActionBarCell`

### DIFF
--- a/app/Entities/Post.php
+++ b/app/Entities/Post.php
@@ -16,7 +16,7 @@ class Post extends Entity
     use HasReactions;
 
     protected $datamap = [];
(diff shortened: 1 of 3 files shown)


neznaika0 (commented):
@lonnieezell @datamweb @michalsn Can you check the PRs? Further development requires a merger, they affect the work of the forum. Maybe I'm in a hurry. I don't see any other way of gradual editing. It would be a pity if we drop the forum again for a year.

datamweb (approved):
LGTM!

lonnieezell (approved):

